PR #77, opened 2026-05-19
fix: move test deps to dependency-groups so uv run pytest works

### LINKED ISSUE #9: Tests: 'uv run pytest' fails silently without --extra dev
## Severity
HIGH (DX / silent failures)

## Finding
Running `uv run pytest` in a fresh checkout produces ~200 failures because `pytest-asyncio` lives in `[project.optional-dependencies] dev`, which `uv run` does not install by default. Every async test reports:
```
async def functions are not natively supported.
```
The actual correct invocation is `uv run --extra dev pytest`. This trips up every new contributor.

## Recommended fix
Either:
1. Move `pytest-asyncio` (and other test deps) into `[dependency-groups] dev` which `uv run` installs by default, OR
2. Add `pytest-asyncio` to base runtime deps (not ideal), OR
3. Document `uv run --extra dev pytest` prominently in README.md, CLAUDE.md, and `contributing/README.md`, AND add a `conftest.py` import-time check that fails loudly with a helpful message.

jwesleye:
Closes #9

## Changes
- Moved `pytest`, `pytest-asyncio`, `pytest-mock`, `httpx`, `ruff`, `mypy`, `black`, `pre-commit` into `[dependency-groups] dev` in `pyproject.toml` — `uv run` includes this group by default, so `uv run pytest` now works without `--extra dev`
- Added import-time guard in `tests/conftest.py` that exits with a clear, actionable error message if `pytest-asyncio` is missing
- Updated `README.md` Development section: install with `uv sync`, run tests with `uv run pytest`
- Updated `CLAUDE.md` Development Setup: replaced `uv pip install -e ".[dev]"` with `uv sync`
- `[project.optional-dependencies] dev` kept intact for backward compatibility with `pip install -e ".[dev]"` users

## Test plan
- `uv sync` installs all dev deps from `[dependency-groups] dev`
- `uv run pytest tests/unit/test_account_tools.py` passes (10 passed, 9 skipped)
- No `--extra dev` flag needed

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -15,8 +15,7 @@ Project guidance for Claude Code when working with the Open Stocks MCP server.
 
 ### Development Setup
 ```bash
-uv venv && source .venv/bin/activate
-uv pip install -e ".[dev]"
+uv sync   # installs project + all dev dependencies
 ```
 
 ### Testing
--- a/README.md
+++ b/README.md
@@ -23,7 +23,7 @@ For development:
 ```bash
 git clone https://github.com/Open-Agent-Tools/open-stocks-mcp.git
 cd open-stocks-mcp
-uv pip install -e .
+uv sync   # installs all deps including dev group
 ```
 
 ## Quick Start
@@ -213,10 +213,10 @@ The server handles Robinhood's authentication requirements:
 
 ### Testing
 ```bash
-pytest                           # All tests
-pytest -m "journey_account"      # Fast account tests (~1.8s)
-pytest -m "journey_market_data"  # Market data tests (~3.8s) 
-pytest -m "not slow and not exception_test"  # Recommended for development
+uv run pytest                           # All tests (works after uv sync)
+uv run pytest -m "journey_account"      # Fast account tests (~1.8s)
+uv run pytest -m "journey_market_data"  # Market data tests (~3.8s)
+uv run pytest -m "not slow and not exception_test"  # Recommended for development
 
 # See CLAUDE.md for complete journey testing guide
 ```
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -165,7 +165,15 @@ managed = true
 
 [dependency-groups]
 dev = [
+    "pytest>=9.0.0",
+    "pytest-asyncio>=1.3.0",
+    "pytest-mock>=3.15.0",
     "pytest-cov>=6.2.1",
+    "httpx>=0.28.0",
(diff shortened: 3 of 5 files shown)


unseriousAI (dismissed):
FOREMAN_CODE_REVIEW
head_sha: d9265d25b5fb75183591a437466bd1649364efcc
status: BLOCK
spec_compliance: BLOCK
blocking_findings: 2
non_blocking_findings: 0

## Summary
REQUEST_CHANGES: the core fix is valid, but the PR also introduces broad, unrelated lockfile dependency churn including unfamiliar packages, which violates scope and dependency safety for this issue.

## Blocking
1. `uv.lock` (multiple sections; e.g. package entries around lines 49-88, 491-506, 662-734): The PR adds unfamiliar dependencies (`ast-serialize`, `librt`) and performs wide transitive/runtime upgrades (for example `fastapi`, `starlette`, `anyio`) that are not required to solve issue #9. This fails the dependency safety check and introduces unvetted supply-chain/runtime risk. Suggested fix: regenerate lockfile in a constrained way so only required test/dev dependency resolution changes are included, or split dependency-upgrade work into a separate vetted PR.
2. `uv.lock` (repo-wide lock delta) vs issue scope `#9`: Issue #9 is specifically about `uv run pytest` failing without extra dev flags. The large lockfile refresh is extra unrelated work and makes reviewability/regression analysis harder than necessary for this bug fix. Suggested fix: reduce this PR to minimal changes needed for test dependency availability and associated docs/conftest updates; move broad package refresh to a dedicated dependency-maintenance PR.

## Suggestions
None

jwesleye:
Foreman unblock update for PR #77:

Addressed blockers:
- Reduced `uv.lock` to the dependency-group metadata needed for this PR and removed the broad unrelated package refresh from the branch diff (`ast-serialize`, `librt`, runtime package upgrades, etc.).
- Removed the unused `# noqa: T201` from the new `tests/conftest.py` import guard so the changed file passes ruff.

Validation evidence:
- `uv run --frozen pytest tests/unit/test_account_tools.py -q` exit 0: 10 passed, 9 skipped.
- `uv run --frozen ruff check tests/conftest.py` exit 0: all checks passed.
- `git diff --check` exit 0.

Note: validation used `--frozen` intentionally because the base lockfile is already stale against current runtime dependency minimums; a normal `uv lock` refreshes unrelated runtime pins, which was the blocker this unblock removes.

PR #78, opened 2026-05-19
fix: block auth retry on persistent cache-clear failures

### LINKED ISSUE #16: Error handling: pickle clear silent failure can cause infinite auth retry
## Severity
MEDIUM

## Location
`src/open_stocks_mcp/tools/session_manager.py:96-98`

## Finding
```python
except Exception as e:
    logger.error(f"Failed to clear pickle file: {e}")
    return False
```
If the pickle clear fails (permission denied, disk full, corrupted file) the auth-retry logic in `execute_with_retry` will keep retrying with the same stale session, possibly infinitely.

## Recommended fix
- Escalate persistent pickle-clear failures (log CRITICAL).
- Short-circuit the retry loop after N consecutive clear failures, returning a structured `AuthenticationError` to the caller.

unseriousAI:
Closes #16

## Changes
- track consecutive Robinhood pickle-clear failures in `SessionManager` and emit `CRITICAL` once threshold is exceeded
- add `should_block_auth_retries()` guard to prevent unsafe re-auth loops when cache clear is persistently failing
- short-circuit `execute_with_retry` auth-retry path with structured `AuthenticationError` when blocker is active
- add focused unit coverage for blocked auth-retry and normal auth-refresh path

## Test plan
- uv run pytest -q tests/unit/test_error_handling.py
- uv run ruff check src/open_stocks_mcp/tools/error_handling.py src/open_stocks_mcp/tools/session_manager.py tests/unit/test_error_handling.py
- uv run mypy src/open_stocks_mcp/tools/error_handling.py src/open_stocks_mcp/tools/session_manager.py

### DIFF
--- a/src/open_stocks_mcp/tools/retry.py
+++ b/src/open_stocks_mcp/tools/retry.py
@@ -70,6 +70,13 @@ async def execute_with_retry(
                 )
 
             if isinstance(classified_error, AuthenticationError) and handle_auth_errors:
+                if session_manager.should_block_auth_retries():
+                    logger.critical(
+                        "Blocking authentication retry due to persistent session cache clear failures"
+                    )
+                    raise AuthenticationError(
+                        "Session cache clear failures prevent safe authentication retry"
+                    ) from e
                 if auth_retry_count < max_auth_retries:
                     logger.warning(
                         f"Authentication error detected, attempting re-authentication: {e}"
--- a/src/open_stocks_mcp/tools/session_manager.py
+++ b/src/open_stocks_mcp/tools/session_manager.py
@@ -30,6 +30,8 @@ class SessionManager:
         self._lock = asyncio.Lock()
         self._is_authenticated = False
         self._failed_login_attempts = 0
+        self._consecutive_pickle_clear_failures = 0
+        self._max_pickle_clear_failures = 3
 
     def set_credentials(self, username: str, password: str) -> None:
         """Store credentials for re-authentication.
@@ -97,13 +99,27 @@ class SessionManager:
             if pickle_path.exists():
                 pickle_path.unlink()
                 logger.info(f"Cleared pickle file: {pickle_path}")
+                self._consecutive_pickle_clear_failures = 0
                 return True
             else:
                 logger.debug(f"Pickle file does not exist: {pickle_path}")
+                self._consecutive_pickle_clear_failures = 0
                 return True
         except Exception as e:
+            self._consecutive_pickle_clear_failures += 1
             logger.error(f"Failed to clear pickle file: {e}")
+            if self._consecutive_pickle_clear_failures >= self._max_pickle_clear_failures:
+                logger.critical(
+                    "Session cache clear failed %s consecutive times; authentication retries will be blocked until cache clear succeeds",
+                    self._consecutive_pickle_clear_failures,
+                )
             return False
+
+    def should_block_auth_retries(self) -> bool:
+        """Return True when persistent pickle clear failures make auth retries unsafe."""
+        return (
+            self._consecutive_pickle_clear_failures >= self._max_pickle_clear_failures
+        )
 
     def _increment_failed_attempts(self) -> None:
         """Increment failed login attempts and clear pickle if threshold reached."""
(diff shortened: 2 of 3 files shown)


jwesleye (dismissed):
FOREMAN_CODE_REVIEW
head_sha: 7e99da4727deba95103e4695338ce714e6b52d48
status: BLOCK
spec_compliance: PASS
blocking_findings: 1
non_blocking_findings: 3

## Summary
Auth-retry circuit breaker is logically sound and addresses issue #16, but `_consecutive_pickle_clear_failures` has no reset path through the normal session lifecycle (`logout()`), leaving users permanently blocked after the underlying filesystem issue is resolved.

## Blocking

1. **`logout()` doesn't reset `_consecutive_pickle_clear_failures`** (`session_manager.py`, `logout()` finally block ~L502–505)

   `logout()` explicitly resets `_failed_login_attempts = 0` but not `_consecutive_pickle_clear_failures`. After the normal flow of: pickle-clear fails 3× → user fixes the FS issue (e.g. clears disk space / corrects permissions) → user calls logout → user attempts an API call → any subsequent auth error in `execute_with_retry` still sees `should_block_auth_retries() == True` and short-circuits. The session is permanently blocked through the standard lifecycle. The only recovery is `force_fresh_login()` or `clear_session_cache()`, which require the user to know to invoke them explicitly via a separate MCP tool call.

   `_failed_login_attempts` and `_consecutive_pickle_clear_failures` are parallel state that track separate failure modes; `logout()` and `set_credentials()` already treat the former as session-scoped state that gets cleared on clean lifecycle transitions. The latter should follow the same rule.

   **Fix:** add `self._consecutive_pickle_clear_failures = 0` in `logout()`'s `finally` block alongside the existing `_failed_login_attempts = 0` reset.

## Suggestions

1. **`_max_pickle_clear_failures` should be a constructor parameter** (`session_manager.py` `__init__` ~L30): hardcoded `3` is inconsistent with `max_failed_attempts` which is already configurable via `__init__`. Making it a param improves unit-test isolation and lets callers tune the threshold independently. A follow-up issue has been filed.

2. **`get_session_info()` doesn't expose blocker state** (~L463–489): `_consecutive_pickle_clear_failures` and `should_block_auth_retries()` are invisible to diagnostics. When the circuit breaker trips, operators have no way to see it via the health/session MCP tools without reading logs. A follow-up issue has been filed to add `"consecutive_pickle_clear_failures"` and `"auth_retries_blocked"` fields.

3. **Rebase required before merge**: PR has a confirmed merge conflict (`CONFLICTING`). PR#97 (merged to main after this branch was cut) refactored `_clear_pickle_file` to also handle the encrypted pickle (`.pickle.enc`) and unified two `return True` branches into one. The counter-reset logic in this PR adds `_consecutive_pickle_clear_failures = 0` in two separate `return True` branches; after rebase that becomes a single reset before the unified `return True`, and the increment/CRITICAL log stays in the `except` block unchanged. Mechanical merge — no logic changes needed.

jwesleye:
Unblocked by Foreman.

**Addressed:**
- Rebased onto main; resolved content conflicts in `error_handling.py` (now a facade — moved the `should_block_auth_retries()` short-circuit into the new `open_stocks_mcp.tools.retry.execute_with_retry` location) and in `tests/unit/test_error_handling.py` (combined main's configurable-retry tests with this PR's auth-retry tests; updated markers to `@pytest.mark.asyncio` for consistency).
- Blocking review finding: `logout()` now resets `_consecutive_pickle_clear_failures = 0` in its `finally` block alongside `_failed_login_attempts = 0`, so the circuit breaker clears through the normal lifecycle.

**Validation:**
- `uv run pytest tests/unit/test_error_handling.py tests/unit/test_session_manager.py tests/unit/test_auth_coordinator.py -v` → 28 passed
- `uv run ruff check src/open_stocks_mcp/tools/error_handling.py src/open_stocks_mcp/tools/retry.py src/open_stocks_mcp/tools/session_manager.py tests/unit/test_error_handling.py` → All checks passed
- `uv run mypy src/open_stocks_mcp/tools/error_handling.py src/open_stocks_mcp/tools/retry.py src/open_stocks_mcp/tools/session_manager.py` → Success: no issues found

Head SHA: `3d47262c2c4d8560ed360dbc3068ed5e67cc6e08`

PR #79, opened 2026-05-19
fix: resolve mypy type safety errors in server/app.py (#7)

### LINKED ISSUE #7: Type safety: 3 mypy 'no-any-return' errors in server/app.py
## Severity
HIGH (type safety; also makes CLAUDE.md's 'zero mypy errors' claim false)

## Location
- `src/open_stocks_mcp/server/app.py:1139`
- `src/open_stocks_mcp/server/app.py:1272`
- `src/open_stocks_mcp/server/app.py:1292`

## Finding
```
error: Returning Any from function declared to return "dict[str, Any]"  [no-any-return]
```
All three return values declared as `dict[str, Any]` but actually return `Any`. Adjacent line 1151 has `# type: ignore[no-any-return]` — these were either added later or were missed.

## Recommended fix
Either:
1. Refactor the call sites to return a properly-typed dict (preferred — `cast(dict[str, Any], ...)` if the call returns a known dict), OR
2. Add `# type: ignore[no-any-return]` on each line matching the pattern used at line 1151.

Then update `CLAUDE.md:129` claim once fixed.

unseriousAI:
Closes #7

## Changes
- Improved typing of `handle_schwab_errors`, `handle_robin_stocks_errors`, and `handle_robin_stocks_sync_errors` decorators in `src/open_stocks_mcp/tools/error_handling.py` using `TypeVar` and `cast` to preserve original function signatures.
- Removed 91 redundant `# type: ignore[no-any-return]` comments in `src/open_stocks_mcp/server/app.py` that were previously required due to erased type information from the decorators.
- Removed redundant `cast(dict[str, Any], ...)` calls in `src/open_stocks_mcp/brokers/robinhood.py`.
- Removed an unused `type: ignore` in `src/open_stocks_mcp/tools/robinhood_watchlist_tools.py`.

## Test plan
- Verified that `mypy .` passes with zero errors (Success: no issues found in 36 source files).
- Verified that unit tests for Schwab market tools still pass (except for a pre-existing `AttributeError` in `test_get_price_history_success` which is also present in the main branch).

### DIFF
--- a/src/open_stocks_mcp/brokers/robinhood.py
+++ b/src/open_stocks_mcp/brokers/robinhood.py
@@ -126,7 +126,7 @@ class RobinhoodBroker(BaseBroker):
         # TODO: Phase 2 - delegate to existing get_account_info() function
         from open_stocks_mcp.tools.robinhood_account_tools import get_account_info
 
-        return cast(dict[str, Any], await get_account_info())
+        return await get_account_info()
 
     async def get_portfolio(self) -> dict[str, Any]:
         """Get portfolio holdings."""
@@ -136,7 +136,7 @@ class RobinhoodBroker(BaseBroker):
         # TODO: Phase 2 - delegate to existing get_portfolio() function
         from open_stocks_mcp.tools.robinhood_account_tools import get_portfolio
 
-        return cast(dict[str, Any], await get_portfolio())
+        return await get_portfolio()
 
     async def get_positions(self) -> dict[str, Any]:
         """Get current positions."""
@@ -146,7 +146,7 @@ class RobinhoodBroker(BaseBroker):
         # TODO: Phase 2 - delegate to existing get_positions() function
         from open_stocks_mcp.tools.robinhood_account_tools import get_positions
 
-        return cast(dict[str, Any], await get_positions())
+        return await get_positions()
 
     async def get_stock_quote(self, symbol: str) -> dict[str, Any]:
         """Get stock quote by symbol."""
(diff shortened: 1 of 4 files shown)


jwesleye (approved):
FOREMAN_CODE_REVIEW
head_sha: ffdfa7180efcd709338b8a3a54043f888c8ed488
status: PASS
spec_compliance: PASS
blocking_findings: 0
non_blocking_findings: 1

## Summary

Root-cause fix for issue #7: TypeVar + cast in the three error-handling decorators eliminates all 91 downstream `# type: ignore[no-any-return]` suppressions throughout `server/app.py`. Clean, idiomatic, and strictly better than the two options the issue suggested.

## Blocking

None

## Suggestions

1. **`error_handling.py` — consider `ParamSpec` for full argument-type preservation.**
   The current `F = TypeVar("F", bound=Callable[..., Any])` preserves call-ability and return type at the cast boundary, but erases parameter types. Decorated functions with typed parameters (e.g., `get_stock_price(symbol: str)`) will still satisfy mypy at the call site, but a future caller doing `handle_robin_stocks_errors(some_typed_fn)` and then calling the result loses arg-type checking. The `ParamSpec`-based alternative:
   ```python
   from typing import ParamSpec, TypeVar
   P = ParamSpec("P")
   R = TypeVar("R")
   def handle_robin_stocks_errors(func: Callable[P, Awaitable[R]]) -> Callable[P, Awaitable[R | dict[str, Any]]]:
   ```
   …would give complete argument and return-type fidelity. Non-blocking — the current approach fully satisfies zero-mypy-errors and is the pattern used by most Python decorator stubs in practice. Worth a follow-up issue if stricter call-site checking is ever desired.

PR #84, opened 2026-05-19
fix: enforce read-only MCP tool authorization

### LINKED ISSUE #38: Security: /mcp endpoint has no per-tool authorization
## Severity
MEDIUM (compounding #2)

## Location
\`src/open_stocks_mcp/server/http_transport.py:296-302\`

## Finding
The \`/mcp\` endpoint calls \`mcp_server.call_tool()\` with user-supplied tool name and args. There is no allow-list, no per-tool permission check. A caller (or anyone reaching the unauthenticated 0.0.0.0 endpoint per #2) can invoke any registered tool — including \`order_buy_market\`, \`order_sell_market\`, etc.

## Recommended fix
Add a per-tool permission model:
- Default \`read-only\` mode: only \`get_*\`, \`list_*\` tools allowed.
- \`--allow-trading\` flag required to enable order placement tools.
- Track a \`requires_auth\` attribute on each tool registration.

unseriousAI:
Closes #38

## Changes
- enforce per-tool authorization for `/mcp` `tools/call` requests in HTTP transport
- default HTTP mode to read-only tool access (`get_*`, `list_*`)
- add `--allow-trading` CLI flag to explicitly allow mutating tools over HTTP
- add focused HTTP transport tests for default deny/allow behavior

## Test plan
- `uv run pytest tests/http_transport/test_http_server.py -q -k "blocks_mutating_tools_by_default or allows_read_only_tools_by_default or allows_mutating_tools_when_enabled"`
- `uv run ruff check src/open_stocks_mcp/server/http_transport.py src/open_stocks_mcp/server/app.py tests/http_transport/test_http_server.py`

## Notes
- Full `tests/http_transport/test_http_server.py` run is currently impacted by existing async fixture/plugin configuration issues in this repo environment; focused authorization tests pass.

### DIFF
--- a/src/open_stocks_mcp/server/app.py
+++ b/src/open_stocks_mcp/server/app.py
@@ -1402,6 +1402,12 @@ def attempt_login(username: str, password: str) -> None:
         "Also read from the MCP_API_KEY environment variable."
     ),
 )
+@click.option(
+    "--allow-trading",
+    is_flag=True,
+    default=False,
+    help="Allow mutating trading tools over HTTP /mcp (default is read-only only)",
+)
 def main(
     port: int,
     host: str,
@@ -1409,6 +1415,7 @@ def main(
     username: str | None,
     password: str | None,
     api_key: str | None,
+    allow_trading: bool,
 ) -> int:
     """Run the server with specified transport and handle authentication.
 
@@ -1457,7 +1464,15 @@ def main(
             logger.info(
                 "Server ready - broker tools available based on authentication status"
             )
-            asyncio.run(run_http_server(server, host, port, api_key=api_key))
+            asyncio.run(
+                run_http_server(
+                    server,
+                    host,
+                    port,
+                    api_key=api_key,
+                    allow_trading=allow_trading,
+                )
+            )
         return 0
     except KeyboardInterrupt:
         logger.info("\nServer stopped by user")
--- a/src/open_stocks_mcp/server/http_transport.py
+++ b/src/open_stocks_mcp/server/http_transport.py
@@ -30,6 +30,95 @@ LOOPBACK_HOSTS = frozenset({"localhost", "127.0.0.1", "::1"})
 # Health/status/root remain open so container orchestrators can probe liveness.
 PROTECTED_PATHS = frozenset({"/mcp", "/sse", "/session/refresh", "/tools"})
 
+READ_ONLY_HTTP_TOOL_NAMES = frozenset(
+    {
+        "account_details",
+        "account_features",
+        "account_info",
+        "account_profile",
+        "account_settings",
+        "aggregate_option_positions",
+        "aggregated_portfolio",
+        "all_option_positions",
+        "all_watchlists",
+        "basic_profile",
+        "broker_status",
+        "build_holdings",
+        "build_user_profile",
+        "complete_profile",
+        "day_trades",
+        "dividends",
+        "dividends_by_instrument",
+        "find_instruments",
+        "find_options",
+        "health_check",
+        "instruments_by_symbols",
+        "interest_payments",
+        "investment_profile",
+        "latest_notification",
+        "list_brokers",
+        "list_tools",
+        "margin_calls",
+        "margin_interest",
+        "market_hours",
+        "metrics_summary",
+        "notifications",
+        "open_option_orders",
+        "open_option_positions",
+        "open_option_positions_with_details",
+        "open_stock_orders",
+        "option_historicals",
+        "option_market_data",
+        "options_chains",
+        "options_orders",
+        "portfolio",
+        "positions",
+        "price_history",
+        "pricebook_by_symbol",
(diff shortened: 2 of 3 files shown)


jwesleye (dismissed):
FOREMAN_CODE_REVIEW
head_sha: 86e3058cd6726504172890dc30cbf317cd01d9ad
status: BLOCK
spec_compliance: BLOCK
blocking_findings: 2
non_blocking_findings: 2

## Summary
The authorization gate denies mutating tools as intended, but the `get_*`/`list_*` prefix filter does not match this codebase's actual registered tool names — so in default read-only mode the HTTP transport would also deny ~80+ legitimate read-only tools, leaving the user only two real options: nothing usable, or `--allow-trading` (which re-enables every mutating tool). This inverts the security intent of #38.

## Blocking

1. **Prefix filter mismatches the real tool registry** — `src/open_stocks_mcp/server/http_transport.py:88-90`
   The filter accepts `tool_name.startswith("get_") or tool_name.startswith("list_")`, but `src/open_stocks_mcp/server/app.py:163-1308` registers FastMCP tools using their bare Python function names (no `name=` override anywhere). The actual registered read-only tool names include `account_info`, `portfolio`, `account_details`, `positions`, `stock_price`, `stock_info`, `market_hours`, `price_history`, `dividends`, `notifications`, `stock_orders`, `options_orders`, `session_status`, `health_check`, `schwab_quote`, `schwab_accounts`, etc. None of these start with `get_` or `list_`. Only `list_tools` and `list_brokers` would pass the filter. As written, the default HTTP mode breaks almost all read-only functionality, and the only way to restore it (`--allow-trading`) re-enables every mutating tool — the exact situation #38 was meant to prevent.

   Suggested fix: either (a) build an explicit allow-list from the actual registered read-only tool names, (b) attach a `mutating: bool` (or `requires_auth: bool`) attribute on each `@mcp.tool()` registration and gate on that, or (c) keep prefix-based gating but rename all read-only tool registrations to use `get_*`/`list_*` (large breaking change to existing MCP clients).

2. **Tests do not exercise real tool names — masks finding #1** — `tests/http_transport/test_http_server.py:19-29, 149-209`
   The fixture was updated to register synthesized tools `get_test_tool` and `order_buy_market` that happen to match the new filter's assumptions. No test verifies that any actually-registered production tool (e.g. `account_info`, `stock_price`, `portfolio`) is permitted in default mode. The tests pass but provide false confidence — they validate the filter in isolation, not against the real registry. Issue #38 even names `order_buy_market` as the canonical example tool, which is what the test uses, but the production registration is `buy_stock_market` (`app.py:809-810`), not `order_buy_market`.

   Suggested fix: add a test that boots `create_http_server` against the real `mcp` instance from `open_stocks_mcp.server.app` (or a sample of its real tool names) and asserts both that representative read-only tools like `account_info` are allowed and that representative mutating tools like `buy_stock_market` are denied.

## Suggestions

1. **#38 explicitly recommended a `requires_auth` attribute per tool registration** — Tracking mutating-ness on the tool itself is more durable than name-based heuristics and survives future renames. Worth reconsidering before settling on prefixes.

2. **JSON-RPC error code `-32600` ("Invalid Request") is semantically off for authorization denial** — `src/open_stocks_mcp/server/http_transport.py:331`. `-32600` is for malformed requests. An implementation-defined code in the server range, or at minimum a message that does not leak which tools exist (the current message reveals the tool name is valid but blocked), would be cleaner. Non-blocking.

unseriousAI:
Foreman unblock completed.

Addressed blockers:
- Rebased `foreman/38-security-mcp-endpoint-has-no-per-tool-authorizatio` onto `origin/main`; merge conflicts in `src/open_stocks_mcp/server/app.py` and `src/open_stocks_mcp/server/http_transport.py` are resolved.
- Preserved the current bearer-token HTTP protection while adding `--allow-trading` as an explicit opt-in for mutating MCP tools.
- Replaced the `get_*` / `list_*` read-only heuristic with an explicit allow-list of real production read-only tool names, so default HTTP mode allows tools such as `account_info`, `portfolio`, `stock_price`, `market_hours`, `schwab_quote`, and `schwab_accounts` while blocking mutating tools such as `buy_stock_market`, `add_to_watchlist`, and `schwab_cancel_order`.
- Updated HTTP transport tests to exercise representative production tool names and verify the allow-list against the production registry sample.

Validation:
- `uv run pytest tests/http_transport/test_http_server.py -q -k "blocks_mutating_tools_by_default or allows_read_only_tools_by_default or allows_mutating_tools_when_enabled or allow_list_matches_production_registry_sample"` -> 4 passed, 15 deselected
- `uv run ruff check src/open_stocks_mcp/server/http_transport.py src/open_stocks_mcp/server/app.py tests/http_transport/test_http_server.py` -> passed
- `git diff --check` -> passed

New head: `de3b852d94bbc9395d85131ede8a0af311c6ef0b`

PR #88, opened 2026-05-19
fix: validate Schwab token_path boundary

### LINKED ISSUE #37: Security: SchwabBroker token_path not validated for directory traversal
## Severity
MEDIUM (potential)

## Location
\`src/open_stocks_mcp/brokers/schwab.py:48-55\`

## Finding
The \`token_path\` parameter is accepted without validation. Currently called only with a hardcoded default, but if it ever flows from config / env / user input, an attacker could write or read arbitrary filesystem paths.

## Recommended fix
\`\`\`python
allowed_root = Path.home() / ".tokens"
resolved = Path(token_path).resolve()
if not resolved.is_relative_to(allowed_root):
    raise ValueError(f"token_path must be under {allowed_root}")
\`\`\`

unseriousAI:
Closes #37

## Changes
- enforce that `SchwabBroker.token_path` resolves under `~/.tokens`
- raise `ValueError` for token paths outside the allowed root
- add unit coverage for accepted and rejected token path cases
- adjust existing Schwab broker fixture to use an allowed token path

## Test plan
- uv run pytest tests/unit/test_schwab_broker.py -k 'token_path or rejects_token_path or accepts_token_path' -q
- uv run ruff check src/open_stocks_mcp/brokers/schwab.py tests/unit/test_schwab_broker.py

### DIFF
--- a/src/open_stocks_mcp/brokers/schwab.py
+++ b/src/open_stocks_mcp/brokers/schwab.py
@@ -44,15 +44,18 @@ class SchwabBroker(BaseBroker):
         self.app_secret = app_secret
         self.callback_url = callback_url
 
+        token_dir = Path.home() / ".tokens"
+        token_dir.mkdir(parents=True, exist_ok=True)
+
         # Default token path
         if token_path is None:
-            token_dir = Path.home() / ".tokens"
-            token_dir.mkdir(exist_ok=True)
             self.token_path = str(token_dir / "schwab_token.json")
         else:
-            self.token_path = token_path
+            self.token_path = str(
+                self._validate_token_path(token_path=token_path, allowed_root=token_dir)
+            )
             # Ensure token directory exists
-            Path(token_path).parent.mkdir(parents=True, exist_ok=True)
+            Path(self.token_path).parent.mkdir(parents=True, exist_ok=True)
 
         self.client = None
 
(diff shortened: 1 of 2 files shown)


jwesleye (approved):
FOREMAN_CODE_REVIEW
head_sha: 24bfc4e4d02dcb2685b5c8386ab2a298b55a21ad
status: PASS
spec_compliance: PASS
blocking_findings: 0
non_blocking_findings: 3

## Summary
Security fix is correct and well-implemented — APPROVE.

## Blocking
None

## Suggestions

1. **Test assertion fragility** (`tests/unit/test_schwab_broker.py`, lines ~150–152): `broker.token_path` stores the *resolved* path (via `_validate_token_path`), while `token_path` in the local variable is unresolved. These match on typical systems but can diverge when the home directory path involves symlinks. Prefer: `assert Path(broker.token_path) == Path(token_path).expanduser().resolve()` to make the test resilient.

2. **Unconditional `~/.tokens` directory creation** (`schwab.py:47–48`): Moving `token_dir.mkdir(parents=True, exist_ok=True)` before the `if token_path is None` branch means the directory is always created, including in tests and dry-run contexts. This is not a bug, but leaves filesystem side-effects in every test that instantiates `SchwabBroker`. Worth noting for future test-isolation improvements.

3. **Missing symlink-escape test**: The implementation correctly uses `.resolve()` on both paths before calling `relative_to()`, which handles a symlink under `~/.tokens/` pointing outside it. Adding a test for this attack vector (e.g., `os.symlink("/etc/passwd", tmp_path / "link.json")` with `token_path` pointing at the symlink) would strengthen the security regression coverage.

PR #89, opened 2026-05-19
security: default HTTP MCP to loopback, require API key for remote bind

### LINKED ISSUE #2: Security: HTTP MCP endpoint binds to 0.0.0.0 with no authentication
## Severity
CRITICAL / HIGH security

## Location
`src/open_stocks_mcp/server/http_transport.py:226`

## Finding
`mcp_server.settings.host = "0.0.0.0"` binds the MCP endpoint to **all** network interfaces with no authentication. Any host reachable on port 3000/3001 can call `/mcp` and invoke any registered tool — including trading orders if a broker is currently authenticated.

## Recommended fix
- Default bind to `127.0.0.1`.
- Require an explicit `--host 0.0.0.0` opt-in flag.
- When bound to a non-loopback interface, require a bearer token / API key on the `/mcp` endpoint.
- Document the security model in `SECURITY.md`.

jwesleye:
Closes #2

## Summary
The HTTP MCP transport previously forced `mcp_server.settings.host = "0.0.0.0"` and exposed `/mcp` with no authentication. Any host reachable on the port could invoke any registered tool — including trading orders against an authenticated broker session.

## Changes
- **Drop the hardcoded `0.0.0.0` override** in `create_http_server` so the caller's host wins.
- **Default `--host` to `127.0.0.1`** (loopback only).
- **Refuse to start** when bound to a non-loopback host without an API key — `run_http_server` raises `RuntimeError` with a clear remediation message.
- **Add `BearerAuthMiddleware`** that requires `Authorization: Bearer <api-key>` on `/mcp`, `/sse`, `/session/refresh`, and `/tools` when an API key is configured. Uses `secrets.compare_digest` for constant-time comparison.
- **Keep `/health`, `/status`, `/`, `/info`** unauthenticated so container orchestrators can probe liveness (the Docker compose healthcheck still works unchanged).
- **`--api-key` flag** on the CLI, also read from the `MCP_API_KEY` env var.
- **Docs**: new "HTTP Transport Security Model" section in `SECURITY.md`; the Docker compose notes now explain that `MCP_API_KEY` is required because the compose file binds to `0.0.0.0`.

## Test plan
- [x] `pytest tests/http_transport/test_http_auth.py -k "Bearer or NonLoopback"` — 9 new tests, all pass:
  - `/mcp`, `/sse`, `/session/refresh`, `/tools` return 401 without a token, 401 with the wrong token, 200 with the right token.
  - `/health` and `/` remain public.
  - `run_http_server(host="0.0.0.0", api_key=None)` raises `RuntimeError("non-loopback…")`.
  - `is_loopback_host` recognizes `127.0.0.1`, `localhost`, `::1` and rejects `0.0.0.0` and a sample LAN IP.
- [x] Full `tests/http_transport/` run shows the same 16 pre-existing failures as `origin/main` — no regressions. (`diff <(grep ^FAILED before) <(grep ^FAILED after)` is empty.)
- [x] `ruff check` and `mypy` clean on changed files.
- [ ] Manual smoke (recommended before merge): start with `open-stocks-mcp-server --transport http --host 0.0.0.0` and confirm it refuses; rerun with `MCP_API_KEY=xxx` and confirm `/mcp` returns 401 without the header and 200 with it.

## Notes for maintainers / Docker users
This is a breaking change for anyone running the HTTP transport on `0.0.0.0` without an API key: the server will now refuse to start. The Docker compose files in `examples/open-stocks-mcp-docker/` were updated to document the requirement, but users will need to add `MCP_API_KEY=<value>` to their `.env` file when they pull this in. Generate one with `python -c "import secrets; print(secrets.token_urlsafe(32))"`.

### DIFF
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -62,6 +62,29 @@ When reporting a vulnerability, please include:
 
 When using Open Stocks MCP:
 
+### HTTP Transport Security Model
+
+The HTTP MCP transport exposes tool invocation over a network socket. Because
+those tools can place trades against an authenticated broker session, the
+endpoint is treated as **trust-equivalent to broker credentials**.
+
+- **Loopback by default**: `--host` defaults to `127.0.0.1`. Only processes
+  running on the same machine can reach `/mcp`.
+- **Explicit opt-in for network exposure**: Binding to a non-loopback address
+  (e.g. `--host 0.0.0.0` for Docker) is allowed only when an API key is set.
+  The server will refuse to start otherwise.
+- **Bearer-token auth on tool endpoints**: When an API key is configured, every
+  request to `/mcp`, `/sse`, `/session/refresh`, and `/tools` must include
+  `Authorization: Bearer <api-key>`. Health and status endpoints stay
+  unauthenticated so container orchestrators can probe liveness.
+- **Set the key via env var or flag**: Pass `--api-key <value>` or set the
+  `MCP_API_KEY` environment variable. Treat the key as a secret with the same
+  care as broker credentials — anyone holding it can invoke trading tools.
+- **TLS termination**: The server does not terminate TLS itself. When exposing
+  the endpoint over an untrusted network, run it behind a reverse proxy
+  (nginx, Caddy, Traefik) that terminates TLS and forwards to the loopback
+  bind, or rely on platform-level encryption (e.g. WireGuard, mTLS sidecar).
+
 ### API Keys and Authentication
 
 - **Environment Variables**: Store all API keys and secrets in environment variables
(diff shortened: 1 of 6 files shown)


unseriousAI (approved):
FOREMAN_CODE_REVIEW
head_sha: 1b7cb7664b9fe5f8f7bc5e4e0f2bb482aa4833ee
status: PASS
spec_compliance: PASS
blocking_findings: 0
non_blocking_findings: 0

## Summary
Approved: implementation matches issue #2 security requirements and adds targeted tests/docs without scope drift.

## Blocking
None

## Suggestions
None

PR #92, opened 2026-05-19
chore: refresh uv lock metadata

### LINKED ISSUE #81: Tooling: uv.lock metadata is stale against pyproject
## Finding

During issue #8, `uv lock --check` reports that `uv.lock` needs to be updated even after the black dependency is removed. Running plain `uv lock` refreshes many unrelated package pins, so that broad lockfile reconciliation was kept out of the formatter cleanup PR.

## Evidence

- `uv lock --check` exits 1 with: `The lockfile at uv.lock needs to be updated, but --check was provided.`
- A plain `uv lock` would update resolver metadata and refresh many unrelated packages, not just remove black.

## Suggested fix

Do a dedicated dependency lock refresh/review so `uv.lock` metadata matches current `pyproject.toml` without mixing it into unrelated tooling fixes.

jwesleye:
Closes #81

## Changes
- regenerated `uv.lock` from current `pyproject.toml` after removing the unparsable stale lockfile
- refreshed lock dependency metadata so dependencies with multiple matches (like `black`) include required source resolution
- kept scope limited to lockfile reconciliation only

## Test plan
- uv lock --check


jwesleye (commented):
FOREMAN_CODE_REVIEW
head_sha: 6289578f9f1c5be419dc784824bf1e6a43c41452
status: PASS
spec_compliance: PASS
blocking_findings: 0
non_blocking_findings: 1

## Summary
Lockfile refresh is clean — all new transitive deps verified against PyPI — and the update actually corrects a pre-existing spec violation (anyio 4.10.0 was pinned below the declared `>=4.12.0` floor).

## Blocking
None

## Suggestions
1. **uv.lock / pyproject.toml version drift prevention** (`uv.lock`) — The old lockfile pinned `anyio==4.10.0` while `pyproject.toml` required `>=4.12.0`. Consider adding a `uv lock --check` step to CI on every PR that touches `pyproject.toml` so the lock never drifts below declared minimums again. A follow-up issue will be filed.

---

**New transitive packages — all independently verified on PyPI:**
- `annotated-doc 0.0.4` — confirmed FastAPI 0.136.1 core dep (`requires_dist: annotated-doc>=0.0.2`)
- `ast-serialize 0.5.0` — mypy internal AST serialization bindings
- `joserfc` — authlib 1.7.2 JWT/JWS dep (confirmed in authlib 1.7.2 metadata)
- `librt` — Linux/non-PyPy platform-specific mypy dep
- `pyjwt[crypto]` — mcp 1.27.1 authentication dep
- `python-discovery 1.3.1` — Python interpreter discovery (virtualenv transitive dep)
- `pytokens` — black 26.5.1 internal dep
- `black 26.5.1` — already declared in dev extras as `black>=25.0.0`

**Note:** GitHub prevents self-review; this COMMENT reflects a PASS verdict that would otherwise be an APPROVE.

PR #264, opened 2026-05-19
chore: cleanup (ruff)

unseriousAI:
## Cleanup summary
- Tool: ruff check --fix (applied lint fixes)
- Tool: ruff format (applied formatting)

## File counts
- src/open_stocks_mcp/brokers/robinhood.py
- tests/unit/test_session_manager.py
- Total files changed: 2

### DIFF
--- a/src/open_stocks_mcp/brokers/robinhood.py
+++ b/src/open_stocks_mcp/brokers/robinhood.py
@@ -1,7 +1,7 @@
 """Robinhood broker implementation using existing robin-stocks integration."""
 
 from datetime import datetime
-from typing import Any, cast
+from typing import Any
 
 from open_stocks_mcp.brokers.base import BaseBroker, BrokerAuthStatus
 from open_stocks_mcp.logging_config import logger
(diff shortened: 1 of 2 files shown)


jwesleye (approved):
FOREMAN_CODE_REVIEW
head_sha: 6bf29b3ef6b298c4227634ce2fe0fbc25e584a7c
status: PASS
spec_compliance: PASS
blocking_findings: 0
non_blocking_findings: 1

## Summary
Clean ruff cleanup: removes one unused import and reformats a with-statement to parenthesized style. No behavioral changes.

## Blocking
None

## Suggestions
1. **tests/unit/test_session_manager.py, line 37–40** — The patch target `"open_stocks_mcp.tools.session_manager.rh.logout"` references a module path that no longer exists after the session manager was moved to `brokers/robinhood_session`. This pre-dates this PR (the diff's before/after both carry the old path), but the test may not be intercepting the real call, rendering it ineffective. Tracked as a follow-up issue.

PR #316, opened 2026-05-20
fix: preserve interactive MFA prompt visibility

### LINKED ISSUE #299: Follow-up: interactive MFA prompt is invisible due to stderr redirect
Identified during review of PR #293.

In `src/open_stocks_mcp/tools/session_manager.py`, `_resolve_mfa_code` writes the user-facing prompt to `sys.stderr` (lines ~252–258). However, the caller `_login_with_device_verification` wraps the login call in:

```python
with redirect_stdout(stdout_buffer), redirect_stderr(stderr_buffer):
    ...
    return self._resolve_mfa_code()
```

The prompt text therefore lands in `stderr_buffer` rather than the actual terminal — the user is asked to enter a verification code but sees no prompt. The env-var path (`ROBINHOOD_MFA_CODE`) works correctly; this only affects the documented interactive fallback.

Fix options:
- Write the prompt to `sys.__stderr__` so it bypasses the redirection.
- Capture and pass through the original stderr file descriptor before `redirect_stderr` is applied.

Also consider adding a unit test that exercises the interactive path (mocked TTY + `sys.stdin`) to lock in the visible-prompt behavior.

unseriousAI:
Closes #299

## Changes
- write interactive MFA prompt output to the original stderr stream (`sys.__stderr__`) with fallback to `sys.stderr`
- add regression test covering `_resolve_mfa_code()` behavior when stderr is redirected
- keep existing env-var MFA and device-approval behavior unchanged

## Test plan
- uv run pytest tests/unit/test_session_manager.py::test_interactive_mfa_prompt_uses_original_stderr_when_stderr_is_redirected -v
- uv run pytest tests/unit/test_session_manager.py -v
- uv run ruff check src/open_stocks_mcp/tools/session_manager.py tests/unit/test_session_manager.py
- uv run ruff format --check src/open_stocks_mcp/tools/session_manager.py tests/unit/test_session_manager.py
- uv run mypy src/open_stocks_mcp/tools/session_manager.py

### DIFF
--- a/src/open_stocks_mcp/tools/session_manager.py
+++ b/src/open_stocks_mcp/tools/session_manager.py
@@ -252,9 +252,10 @@ class SessionManager:
         if sys.stdin.isatty():
             try:
                 # Use stderr for prompt to avoid polluting stdout in some environments
-                sys.stderr.write("\nROBINHOOD MFA REQUIRED\n")
-                sys.stderr.write("Enter verification code: ")
-                sys.stderr.flush()
+                prompt_stream = getattr(sys, "__stderr__", None) or sys.stderr
+                prompt_stream.write("\nROBINHOOD MFA REQUIRED\n")
+                prompt_stream.write("Enter verification code: ")
+                prompt_stream.flush()
                 return sys.stdin.readline().strip()
             except Exception as e:
                 logger.error(f"Failed to read interactive MFA code: {e}")
--- a/tests/unit/test_session_manager.py
+++ b/tests/unit/test_session_manager.py
@@ -1,6 +1,8 @@
 """Unit tests for session manager behavior."""
 
 import asyncio
+import io
+from contextlib import redirect_stderr
 from datetime import datetime
 from unittest.mock import patch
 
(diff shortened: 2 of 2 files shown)


jwesleye (approved):
FOREMAN_CODE_REVIEW
head_sha: 36d4d05258bbd3d59d6a03d6f3f1b86233f11e9f
status: PASS
spec_compliance: PASS
blocking_findings: 0
non_blocking_findings: 1

## Summary

Clean, minimal fix that correctly routes the interactive MFA prompt to the original stderr stream, bypassing the `redirect_stderr` capture in `_login_with_device_verification`.

## Blocking

None

## Suggestions

1. **tests/unit/test_session_manager.py** (line 129) — The negative assertion `assert "ROBINHOOD MFA REQUIRED" not in redirected_stderr.getvalue()` would be slightly stronger as `assert redirected_stderr.getvalue() == ""` to confirm nothing leaked at all, not just the known string. Low priority — current assertion fully covers the reported bug.

---

**Hard checks**: diff_size=pass (34 lines), credentials=pass, dangerous_patterns=pass, dependencies=pass

**Spec compliance**: Issue #299 requested (a) writing to `sys.__stderr__` to bypass redirection, and (b) adding a unit test for the interactive path. Both are present and correct.

**Code quality**: `getattr(sys, "__stderr__", None) or sys.stderr` is the right idiom — it handles the rare case where `__stderr__` is None and degrades gracefully. The test patches `sys.__stderr__` and uses `redirect_stderr` in the same context block as the call under test, so it faithfully reproduces the production scenario described in the issue.

PR #419, opened 2026-05-21
feat: centralize broker request timeout/retry configuration

### LINKED ISSUE #143: Feature: implement Phase 8 caching and request optimization: Centralize broker request timeout/retry configuration
Parent: #136
depends on #142

## End-to-end behavior delivered by this slice
Unify timeout and retry policy configuration for broker HTTP calls so every broker tool uses consistent defaults and operators can adjust reliability/performance without touching call sites.

## Acceptance Criteria
- [ ] Timeout and retry settings are defined in a central configuration surface.
- [ ] Robinhood and Schwab request paths consume centralized settings.
- [ ] Existing behavior remains backward compatible by default values.

## Likely Files
- src/open_stocks_mcp/brokers/
- src/open_stocks_mcp/tools/error_handling.py
- src/open_stocks_mcp/config/

## Tests
- Unit tests for configured timeout/retry policy application.
- Regression tests confirming retry/error mapping behavior remains stable.

## Validation
- uv run pytest tests/unit -k 'retry or timeout or broker'
- uv run ruff check src tests
- uv run mypy src

unseriousAI:
Closes #143

## Changes
- Centralized timeout and retry configuration in `src/open_stocks_mcp/config.py`.
- Implemented `install_robinhood_request_timeout` in `src/open_stocks_mcp/brokers/request_policy.py` to monkeypatch `robin_stocks` session.
- Implemented `execute_broker_request` shared async executor with retry and deadline logic.
- Updated `RobinhoodBroker` and `SchwabBroker` to consume centralized timeout settings.
- Refactored all Schwab tools (`account`, `market`, `options`, `trading`) to use the shared executor.
- Ensured non-idempotent operations (orders, cancellations) do not retry.
- Added comprehensive unit tests for the new request policy logic.
- Updated existing Schwab tool tests to verify usage of the new executor.

## Test plan
- `uv run pytest tests/unit/test_broker_request_policy.py` - new policy tests.
- `uv run pytest tests/unit/test_schwab_*.py` - updated Schwab tests.
- Verified zero MyPy errors and Ruff linting passed.

### DIFF
--- a/src/open_stocks_mcp/brokers/request_policy.py
+++ b/src/open_stocks_mcp/brokers/request_policy.py
@@ -1,0 +1,117 @@
+import asyncio
+import functools
+import time
+from typing import Any, Callable, TypeVar
+
+from open_stocks_mcp.logging_config import logger
+
+T = TypeVar("T")
+
+def install_robinhood_request_timeout(timeout_seconds: float, session: Any | None = None) -> None:
+    """Idempotently install a timeout policy on a requests-like session.
+    
+    This wraps the session's request method to ensure every call uses the
+    centrally configured timeout, overriding any call-site defaults provided
+    by the SDK.
+    """
+    if session is None:
+        try:
+            from robin_stocks.robinhood import helper
+            session = helper.SESSION
+        except ImportError:
+            logger.warning("robin_stocks not installed; skipping Robinhood timeout policy installation")
+            return
+
+    # Check if already installed
+    if getattr(session.request, "_is_timeout_wrapper", False) is True:
+        session._robinhood_timeout = timeout_seconds
+        return
+
+    session._robinhood_timeout = timeout_seconds
+    original_request = session.request
+
+    @functools.wraps(original_request)
+    def timeout_wrapper(*args: Any, **kwargs: Any) -> Any:
+        kwargs["timeout"] = getattr(session, "_robinhood_timeout", 16.0)
+        return original_request(*args, **kwargs)
+
+    setattr(timeout_wrapper, "_is_timeout_wrapper", True)
+    session.request = timeout_wrapper
+    logger.debug(f"Installed Robinhood request timeout policy: {timeout_seconds}s")
+
+
+async def execute_broker_request(
+    func: Callable[..., T],
+    *args: Any,
+    policy: Any | None = None,
+    retry_safe: bool = True,
+    **kwargs: Any,
+) -> T:
+    """Execute a synchronous broker SDK call with optional retries and deadline enforcement.
+
+    Args:
+        func: The synchronous broker SDK function to call.
+        policy: Optional BrokerRequestConfig override.
+        retry_safe: Whether it's safe to retry this operation (True for reads, False for mutations).
+        *args, **kwargs: Arguments passed to the SDK function.
+    """
+    if policy is None:
+        from open_stocks_mcp.config import get_config
+
+        policy = get_config().broker_requests
+
+    # max_retries: int
+    # initial_delay: float
+    # backoff_factor: float
+    # deadline: float | None
+
+    max_retries = int(policy.retry_max_retries) if retry_safe else 0
+    initial_delay = float(policy.retry_initial_delay)
+    backoff_factor = float(policy.retry_backoff_factor)
+    deadline = policy.total_deadline_seconds
+    if deadline is not None:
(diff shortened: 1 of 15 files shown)


jwesleye (dismissed):
FOREMAN_CODE_REVIEW
head_sha: 8319eecf9587190a1c31405594de3350dbf011f5
status: BLOCK
spec_compliance: PASS
blocking_findings: 1
non_blocking_findings: 4

## Summary

Good centralization work — the new `execute_broker_request` / `install_robinhood_request_timeout` design is clean and the production code correctly passes `retry_safe=False` for all order/cancel operations. One blocking gap: the updated tests mock `execute_broker_request` but never assert the `retry_safe` argument, so the critical no-retry safety property for trading mutations is untested and regression-prone.

## Blocking

1. **Missing `retry_safe=False` assertion in trading/cancel tests** — `tests/unit/test_schwab_trading_tools.py`, `tests/unit/test_schwab_options_tools.py`

   The PR's central safety contract is that order placements and cancellations must *never* retry (`retry_safe=False`). The updated tests replace the `asyncio.to_thread` mock with `execute_broker_request` but carry over the old call pattern: mock is set up with a return value and then verified only that it was called. The `retry_safe` kwarg is never asserted.

   If a future change accidentally flips `retry_safe=True` on any trading tool, tests will still pass — but the live behavior could place duplicate orders.

   Required fix: for every test that covers a buy, sell, limit, or cancel operation, add an assertion like:
   ```python
   _, kwargs = mock_to_thread.call_args
   assert kwargs.get("retry_safe") is False
   ```
   or use `mock_to_thread.assert_called_once_with(<func>, retry_safe=False)`.

## Suggestions

1. **`BrokerRequestConfig` duplicates `RetryConfig` env vars** — `src/open_stocks_mcp/config.py` lines 227–242

   `OPEN_STOCKS_MCP_RETRY_MAX_RETRIES`, `OPEN_STOCKS_MCP_RETRY_INITIAL_DELAY`, and `OPEN_STOCKS_MCP_RETRY_BACKOFF_FACTOR` now populate both `RetryConfig` and `BrokerRequestConfig`. Operators setting those vars get the same values in both objects, but the two configs are separate — a confusing state for the goal of centralization. Consider having `BrokerRequestConfig` read from `RetryConfig` as its defaults, or give the broker-specific retry fields distinct env var names.

2. **`assert self.client is not None` in authenticate paths** — `src/open_stocks_mcp/brokers/schwab.py` lines 165, 176

   `assert` is elided under `python -O`. Use an explicit `if self.client is None: raise RuntimeError(...)` guard if this invariant matters at runtime.

3. **Unnecessary intermediate variable in `broker_utils.py`** — `src/open_stocks_mcp/tools/broker_utils.py` lines 265–268

   ```python
   result: tuple[Any, dict[str, Any] | None] = registry.get_broker_or_error(...)
   return result
   ```
   The annotation adds no value mypy can't infer. One-liner `return registry.get_broker_or_error(broker_name, operation)` is cleaner.

4. **Dead code after retry loop** — `src/open_stocks_mcp/brokers/request_policy.py` lines 121–123

   The `for` loop always exits via `return` (success) or `raise` (max retries / deadline). The `if last_exception: raise last_exception` block and `raise RuntimeError("Unreachable")` are never reached. The `RuntimeError` sentinel is a reasonable safety net, but `if last_exception: raise last_exception` above it is pure dead code that can be removed.

unseriousAI:
Foreman unblock pass for PR #419.

Addressed blockers:
- Rebased `foreman/143-centralize-broker-request-config` onto `origin/main`, resolving content conflicts in config, Schwab trading, and Schwab tests.
- Added `retry_safe=False` assertions for Schwab trading/order/cancel option mutation tests, including failure-path coverage.

Validation:
- `uv run pytest tests/unit/test_broker_request_policy.py tests/unit/test_schwab_trading_tools.py tests/unit/test_schwab_options_tools.py tests/unit/test_schwab_market_tools.py -q` -> 55 passed, 3 warnings.
- `uv run ruff check src/open_stocks_mcp/config.py src/open_stocks_mcp/tools/schwab_trading_tools.py tests/unit/test_schwab_trading_tools.py tests/unit/test_schwab_options_tools.py tests/unit/test_schwab_market_tools.py` -> All checks passed.
- `git diff --check` -> passed.

PR #467, opened 2026-05-21
feat: add rate-limit queueing and symbol batching

### LINKED ISSUE #168: Feature: implement Phase 8 caching and request optimization: intelligent rate-limit batching and request optimization
Parent: #136

Adds request batching and rate-limit-aware scheduling so that bursts of related tool calls are coalesced into fewer API requests, and queued calls are smoothed across the rate-limit window.

## Acceptance Criteria
- [ ] Batch helper groups multiple symbol lookups into a single Robin Stocks API call where the API supports it
- [ ] Rate-limit scheduler queues excess requests and drains them within the allowed window rather than dropping them
- [ ] Batching behavior is opt-in per tool and covered by unit tests
- [ ] Rate-limit batching configuration (batch_size, queue_max_wait) is exposed in `config.py`

## Likely Files
- `src/open_stocks_mcp/rate_limiter.py` (modify — add queue/drain logic)
- `src/open_stocks_mcp/config.py` (modify — BatchConfig)
- One or more tool files under `src/open_stocks_mcp/tools/` (adopt batching)
- `tests/unit/test_rate_limiter.py` (modify or create)

## Tests
- Verify batched symbol list produces one API call not N
- Verify queued requests complete within configured max_wait

## Validation
```
pytest tests/unit/test_rate_limiter.py -v
ruff check . --fix && ruff format .
mypy .
```

depends on #159

unseriousAI:
Closes #168

## Changes
- added `BatchConfig` (`batch_size`, `queue_max_wait`) to runtime config with env parsing
- upgraded `RateLimiter.acquire()` to FIFO queued draining with optional `max_wait`, preserving stats/reset behavior
- added reusable `batch_fetch_symbols()` helper with per-loop/per-batch isolation and reset support
- wired stock instrument lookups (`get_stock_info`, `get_instruments_by_symbols`) to opt-in batching while preserving response schema
- added focused unit coverage for queue draining, config parsing, batching coalescing, and concurrent instrument lookups

## Test plan
- `uv run pytest tests/unit/test_rate_limiter.py tests/unit/test_simple_rate_limiter.py -q`
- `uv run pytest tests/unit/test_stock_market_tools.py -q -k 'stock_info or instruments or batch'`
- `uv run pytest tests/unit/ -m 'not slow and not exception_test' -q`
- `uv run ruff check src/open_stocks_mcp/config.py src/open_stocks_mcp/tools/rate_limiter.py src/open_stocks_mcp/tools/robinhood_stock_tools.py tests/conftest.py tests/unit/test_rate_limiter.py tests/unit/test_simple_rate_limiter.py tests/unit/test_stock_market_tools.py`
- `uv run ruff format --check src/open_stocks_mcp/config.py src/open_stocks_mcp/tools/rate_limiter.py src/open_stocks_mcp/tools/robinhood_stock_tools.py tests/conftest.py tests/unit/test_rate_limiter.py tests/unit/test_simple_rate_limiter.py tests/unit/test_stock_market_tools.py`
- `uv run mypy src/open_stocks_mcp/config.py src/open_stocks_mcp/tools/rate_limiter.py src/open_stocks_mcp/tools/robinhood_stock_tools.py`

### DIFF
--- a/src/open_stocks_mcp/config.py
+++ b/src/open_stocks_mcp/config.py
@@ -158,6 +158,14 @@ class OtelConfig:
 
 
 @dataclass
+class BatchConfig:
+    """Configuration for request batching and queue wait behavior."""
+
+    batch_size: int = 10
+    queue_max_wait: float = 0.5
+
+
+@dataclass
 class ServerConfig:
     """Configuration for the MCP server"""
 
@@ -170,6 +178,7 @@ class ServerConfig:
     circuit_breaker: CircuitBreakerConfig = field(default_factory=CircuitBreakerConfig)
     broker_requests: BrokerRequestConfig = field(default_factory=BrokerRequestConfig)
     otel: OtelConfig = field(default_factory=OtelConfig)
+    batch: BatchConfig = field(default_factory=BatchConfig)
 
     def __post_init__(self) -> None:
         if self.retry is None:
@@ -257,6 +266,10 @@ def load_config() -> ServerConfig:
             enabled=otel_enabled,
             service_name=os.getenv("OTEL_SERVICE_NAME", "open-stocks-mcp"),
             exporter_endpoint=os.getenv("OTEL_EXPORTER_OTLP_ENDPOINT") or None,
+        ),
+        batch=BatchConfig(
+            batch_size=_env_int("OPEN_STOCKS_MCP_BATCH_SIZE", 10),
+            queue_max_wait=_env_float("OPEN_STOCKS_MCP_QUEUE_MAX_WAIT", 0.5),
         ),
     )
 
--- a/src/open_stocks_mcp/tools/rate_limiter.py
+++ b/src/open_stocks_mcp/tools/rate_limiter.py
@@ -3,9 +3,11 @@
 import asyncio
 import time
 from collections import defaultdict, deque
+from collections.abc import Awaitable, Callable
 from typing import Any
 
 from open_stocks_mcp.logging_config import logger
+from open_stocks_mcp.tools.exceptions import RateLimitError
 
 
 class RateLimiter:
@@ -39,12 +41,14 @@ class RateLimiter:
         self.broker_buckets: dict[str, deque[float]] = defaultdict(
             lambda: deque(maxlen=100)
         )
+        self._wait_queue: deque[object] = deque()
 
     async def acquire(
         self,
         endpoint: str | None = None,
         weight: float = 1.0,
         broker: str | None = None,
+        max_wait: float | None = None,
     ) -> None:
         """Acquire permission to make an API call.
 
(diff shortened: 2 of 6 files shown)


jwesleye (changes requested):
FOREMAN_CODE_REVIEW
head_sha: e5c3867d3544ca835031f68bc0a6732b11341837
status: BLOCK
spec_compliance: PASS
blocking_findings: 2
non_blocking_findings: 5

## Summary
Spec acceptance criteria all satisfied, but the queueing/batching primitives have two reliability bugs that can permanently hang downstream callers in production.

## Blocking

1. **`RateLimiter.acquire()` leaks waiters on cancellation** — `src/open_stocks_mcp/tools/rate_limiter.py:61-98`
   The function appends `token` to `_wait_queue` but the `while True:` loop has no `try/finally` to remove it. If the caller is cancelled while awaiting `asyncio.sleep(wait_time)` (server shutdown, request timeout, `asyncio.wait_for`, etc.), the token stays at the head of the queue forever. Subsequent `acquire()` calls enqueue their own tokens, then enter the loop and check `if self._wait_queue and self._wait_queue[0] is token:` — that check is always `False` for them because the orphaned token is at the head, so they never enter the capacity check and never make progress. One cancelled caller → the whole queue is wedged until `reset()` is called manually. Production tools call `acquire()` without `max_wait`, so there is no automatic recovery.
   *Suggested fix:* wrap the wait loop in `try: ... finally: self._remove_waiter(token)` so the token is removed on every exit path, including `CancelledError`.

2. **`_drain_batches` orphans futures when `fetch_many` raises** — `src/open_stocks_mcp/tools/rate_limiter.py:281-307`
   Inside the loop, after popping `chunk_symbols` from `state.pending`, the code calls `rows = await fetch_many(chunk_symbols)`. If `fetch_many` raises (network error, auth failure, rate-limit exception from upstream — all common for the wrapped `rh.get_instruments_by_symbols`), the exception propagates out of the loop, the `finally` clears `drain_task`, and every future in `chunk` is left unresolved. Every caller in `batch_fetch_symbols` that is `await fut`-ing one of those futures hangs forever. The PR adds no test for this path.
   *Suggested fix:* wrap `fetch_many` and the result-distribution block in `try/except`, and on exception call `fut.set_exception(exc)` for every future in `chunk` so awaiters surface the failure. Also resolve any leftover symbols on the way out of the loop (e.g. in the finally) so a panic during distribution does not strand them either.

## Suggestions

1. **`test_queue_drains_fifo` does not test what its name claims** — `tests/unit/test_rate_limiter.py:18-31`
   The test mocks `asyncio.sleep` to a no-op and only asserts `len(limiter.call_times) == 2`. It never verifies FIFO ordering between competing acquirers, queue head-of-line behavior, or that the second caller actually waited. Recommend adding a real concurrency test (two `asyncio.gather`-ed acquirers, assert order of completion) and a test for the `max_wait` deadline raising `RateLimitError`.

2. **Drain-end / new-call race in `_drain_batches`** — `src/open_stocks_mcp/tools/rate_limiter.py:281-307`
   Between `if not state.pending: return` and the `finally` block setting `state.drain_task = None`, a new caller can acquire `state.lock`, see `drain_task` still set (not done yet — finally hasn't run), append a symbol to `state.pending` without scheduling a new drain task. Then `finally` runs, clears `drain_task`, and the just-added symbol's future is orphaned until another unrelated caller arrives. Low probability with the default 0.5s `queue_max_wait`, but exercisable with the 0.01s default the new test uses. Consider re-checking `state.pending` inside the finally and rescheduling if non-empty.

3. **Dead filter in `batch_fetch_symbols`** — `src/open_stocks_mcp/tools/rate_limiter.py:265-269`
   `if fut is not None` can never be false — futures returned by `loop.create_future()` are always concrete objects. Either remove the filter or document the invariant.

4. **`_fetch_instruments_batched` silently drops None rows** — `src/open_stocks_mcp/tools/robinhood_stock_tools.py:40`
   The pre-PR `get_instruments_by_symbols` returned whatever the API returned (including None entries for unknown symbols). The new wrapper filters Nones, so callers can no longer tell which symbols failed lookup. For `get_stock_info(symbol)` this is fine (single-symbol path); for the multi-symbol `get_instruments_by_symbols` tool it changes the response semantics — `requested_symbols` and `instruments` may now have different lengths without any explanation. Consider preserving alignment or annotating the missing symbols.

5. **`request_weight = max(1, int(weight))` changes behavior for fractional weights** — `src/open_stocks_mcp/tools/rate_limiter.py:59`
   The old code did `for _ in range(int(weight))`, so a `weight=0.5` call recorded zero call_times entries. The new code floors fractional weights up to 1. No current caller passes a sub-1 weight, but the semantic change is worth noting and probably worth a comment so the next caller is not surprised.

For each finding above: file path, line range, and a suggested fix are inline.

PR #643, opened 2026-05-26
fix: wrap MonitoredTool exceptions to preserve chain (#20)

### LINKED ISSUE #20: Error handling: monitoring.py uses bare 'raise' losing exception chain
## Severity
LOW

## Location
`src/open_stocks_mcp/monitoring.py:279-281`

## Finding
```python
except Exception as e:
    error_type = type(e).__name__
    raise
```
The captured exception variable is discarded; the bare \`raise\` is technically correct but loses the explicit chain when wrapped.

## Recommended fix
Either drop the unused \`error_type\` line, or — if a wrapping exception is intended — use \`raise SomeError(...) from e\`.

unseriousAI:
Closes #20

## Changes
- Replaced bare `raise` in `MonitoredTool.wrapper` with explicit exception chaining using `RuntimeError`.
- Preserved original exception as `__cause__` of the wrapping exception.
- Added regression test in `tests/unit/test_monitoring.py` to verify exception wrapping and metrics recording.

## Test plan
- Run focused regression test: `uv run pytest tests/unit/test_monitoring.py::test_monitored_tool_wraps_exception_with_explicit_cause_and_records_metric`
- Run all monitoring tests: `uv run pytest tests/unit/test_monitoring.py`
- Verified with `ruff` and `mypy`.

### DIFF
--- a/src/open_stocks_mcp/monitoring.py
+++ b/src/open_stocks_mcp/monitoring.py
@@ -674,7 +674,7 @@ class MonitoredTool:
 
             except Exception as e:
                 error_type = type(e).__name__
-                raise
+                raise RuntimeError(f"Monitored tool {self.tool_name} failed") from e
 
             finally:
                 duration = time.time() - start_time
--- a/tests/unit/test_monitoring.py
+++ b/tests/unit/test_monitoring.py
@@ -5,7 +5,7 @@ from unittest.mock import AsyncMock
 
 import pytest
 
-from open_stocks_mcp.monitoring import MetricsCollector
+from open_stocks_mcp.monitoring import MetricsCollector, MonitoredTool
 
 
 @pytest.mark.asyncio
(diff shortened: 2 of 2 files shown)


jwesleye (approved):
FOREMAN_CODE_REVIEW
head_sha: 6761dd0cef2f6fb0a1e08373ca8f4e97424c9d9e
status: PASS
spec_compliance: PASS
blocking_findings: 0
non_blocking_findings: 1

## Summary
Clean, minimal fix: bare `raise` replaced with explicit exception chaining that adds tool-name context without losing the original cause.

## Blocking
None

## Suggestions
1. **`src/open_stocks_mcp/monitoring.py:677` — behavior contract note** — The change shifts the raised type from the original exception (e.g. `ValueError`) to `RuntimeError`. `MonitoredTool` is already marked deprecated and has zero callers in `src/`, so blast radius is nil today. If the deprecation notice is ever lifted or callers are added, document this wrapping contract explicitly on the class so future adopters know to catch `RuntimeError` (or `__cause__`) rather than the specific underlying type.

PR #646, opened 2026-05-26
test: strengthen weak assertions in account, analytics, and auth tests

### LINKED ISSUE #26: Tests: weak assertions in existing test files
## Severity
LOW

## Locations
- \`tests/unit/test_account_tools.py:76, 109, 128, 187\`
- \`tests/unit/test_analytics_tools.py:82, 97\`
- \`tests/unit/test_auth_coordinator.py:330, 346\`

## Finding
Assertions like \`assert isinstance(result["result"], dict)\` and \`assert error is not None\` verify only the response shape, not the actual values. These don't catch regressions where the wrong data is returned with the right shape.

## Recommended fix
Replace with value-level assertions:
\`\`\`python
assert result["result"]["status"] == "success"
assert result["result"]["holdings"][0]["symbol"] == "AAPL"
\`\`\`

unseriousAI:
Closes #26

## Changes
- Hardened assertions in `tests/unit/test_account_tools.py` to verify exact portfolio, positions, and account detail values.
- Strengthened no-data assertions in `tests/unit/test_analytics_tools.py` to check the complete response payload.
- Replaced non-null error checks in `tests/unit/test_auth_coordinator.py` with exact structured error response assertions.

## Test plan
- Run `uv run pytest tests/unit/test_account_tools.py tests/unit/test_analytics_tools.py tests/unit/test_auth_coordinator.py -m "not slow and not exception_test" -q`
- Verified that hardened tests pass and accurately catch regressions in response values.

### DIFF
--- a/tests/unit/test_account_tools.py
+++ b/tests/unit/test_account_tools.py
@@ -81,9 +81,12 @@ class TestAccountTools:
 
         result = await get_portfolio()
 
-        assert "result" in result
-        # The actual response structure depends on how the tool processes the data
-        assert isinstance(result["result"], dict)
+        assert result["result"] == {
+            "market_value": "2000.00",
+            "equity": "2100.00",
+            "buying_power": "100.00",
+            "status": "success",
+        }
 
     @pytest.mark.journey_portfolio
     @pytest.mark.unit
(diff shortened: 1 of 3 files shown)


jwesleye (approved):
FOREMAN_CODE_REVIEW
head_sha: 4ccf1d9c01d872dd19d1f2bb63346fc5cf31ae45
status: PASS
spec_compliance: PASS
blocking_findings: 0
non_blocking_findings: 1

## Summary

Assertion hardening is correct, well-scoped to the linked issue, and verified against implementation. APPROVE.

## Blocking

None

## Suggestions

1. **test_analytics_tools.py — `test_build_holdings_no_data` is a skipped test** (lines 67–84): The strengthened assertion is correct, but this test is marked `@pytest.mark.skip` and `@pytest.mark.exception_test`, so it will not run in normal CI. The improved assertion is safe to keep but provides no regression protection until someone explicitly runs exception tests. Consider removing the skip or converting to a standard unit test if no-data behavior is important to guard.

PR #707, opened 2026-05-26
refactor: use ParamSpec in error-handling decorators for full arg-type preservation

### LINKED ISSUE #91: Follow-up: use ParamSpec in error-handling decorators for full arg-type preservation
Identified during review of PR #79.

The current `TypeVar`-based implementation of `handle_robin_stocks_errors`, `handle_robin_stocks_sync_errors`, and `handle_schwab_errors` in `src/open_stocks_mcp/tools/error_handling.py` correctly satisfies zero-mypy-errors but erases parameter types on decorated functions.

A `ParamSpec`-based approach would preserve both argument types and return types end-to-end:

```python
from typing import ParamSpec, TypeVar, Awaitable, Callable
import functools

P = ParamSpec("P")
R = TypeVar("R")

def handle_robin_stocks_errors(
    func: Callable[P, Awaitable[R]]
) -> Callable[P, Awaitable[R | dict[str, Any]]]:
    @functools.wraps(func)
    async def wrapper(*args: P.args, **kwargs: P.kwargs) -> R | dict[str, Any]:
        context = f"in {func.__name__}"
        try:
            return await func(*args, **kwargs)
        except Exception as e:
            return create_error_response(e, context)
    return wrapper
```

This would give callers full argument-type checking on decorated tools rather than falling back to `(*args: Any, **kwargs: Any)`.

Non-blocking for the current release — the existing cast approach is idiomatic and widely used. This is a quality-of-life improvement for future contributors.

unseriousAI:
Closes #91

## Changes
- Replaced TypeVar + cast with ParamSpec in handle_robin_stocks_errors, handle_schwab_errors, and added handle_robin_stocks_sync_errors in src/open_stocks_mcp/tools/responses.py.
- Preserved argument and return types end-to-end for better mypy checking on decorated tools.
- Updated src/open_stocks_mcp/tools/error_handling.py facade to include the new sync decorator.
- Added regression tests in tests/unit/test_error_handling.py to verify typed parameter preservation at runtime.

## Test plan
- uv run mypy src/ passes with zero errors.
- uv run pytest tests/unit/test_error_handling.py -q passes (including new regression tests).
- uv run ruff check src/open_stocks_mcp/tools/responses.py passes.


jwesleye (approved):
FOREMAN_CODE_REVIEW
head_sha: 8267b36f11f9f779a9d49d2c90bb81380fa11510
status: PASS
spec_compliance: PASS
blocking_findings: 0
non_blocking_findings: 2

## Summary
Clean ParamSpec refactor that correctly replaces TypeVar+cast with end-to-end type preservation. All 96 tests pass, mypy clean.

## Blocking
None

## Suggestions
1. `tests/unit/test_error_handling.py:256` — `handle_robin_stocks_sync_errors` is imported inline within the test method rather than at the module-level import block (line 16-34) where the other decorators are imported. Minor inconsistency.

2. `src/open_stocks_mcp/tools/responses.py:51-64` — `handle_robin_stocks_sync_errors` is newly introduced but has no callers in the codebase yet. This is fine as a completeness addition per issue #91, but worth noting for future cleanup if it remains unused.

PR #708, opened 2026-05-26
docs: update logout() docstring to document fail-hard contract

### LINKED ISSUE #90: Follow-up: update logout() docstring to document fail-hard contract
Identified during review of PR #76.

The `SessionManager.logout()` docstring (session_manager.py:407) still reads just "Logout and clear session." after the exception-propagation change in #76.

Issue #15 explicitly required documenting the contract. The docstring should state that exceptions are logged then re-raised, and that session state is always reset via the finally block even when an exception occurs.

unseriousAI:
Closes #90

## Changes
- Updated `SessionManager.logout()` docstring in `src/open_stocks_mcp/brokers/session_state.py` to explicitly document that exceptions are logged and re-raised, and that session state is always reset via the `finally` block.
- Added a focused unit test in `tests/unit/test_session_manager.py` to assert the presence of these contract terms in the docstring.

## Test plan
- Run `uv run pytest tests/unit/test_session_manager.py -k 'logout' -q` to verify that the docstring contract test and existing behavior tests pass.

### DIFF
--- a/src/open_stocks_mcp/brokers/session_state.py
+++ b/src/open_stocks_mcp/brokers/session_state.py
@@ -216,6 +216,11 @@ class SessionManager:
         return info
 
     async def logout(self) -> None:
+        """Logout and clear session.
+
+        Robin Stocks logout exceptions are logged and re-raised. Local
+        session state is reset in the finally block even when logout raises.
+        """
         async with self._lock:
             try:
                 loop = asyncio.get_event_loop()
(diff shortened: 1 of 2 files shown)


jwesleye (approved):
FOREMAN_CODE_REVIEW
head_sha: d66032bf748c99971c18f91850084737f4c7a4f5
status: PASS
spec_compliance: PASS
blocking_findings: 0
non_blocking_findings: 0

## Summary
Clean, minimal documentation fix — docstring accurately describes the fail-hard contract and tests verify the terms are present.

## Blocking
None

## Suggestions
None

PR #712, opened 2026-05-26
fix: use proper JSON-RPC error code for tool authorization denial

### LINKED ISSUE #94: Follow-up: use proper JSON-RPC error code for tool authorization denial
Identified during review of PR #84.

The new authorization gate in `src/open_stocks_mcp/server/http_transport.py` returns JSON-RPC error code `-32600` ("Invalid Request") when a tool call is blocked in read-only mode. `-32600` is intended for malformed requests, not authorization failures.

Suggested options:
- Use an implementation-defined code in the server range (`-32000` to `-32099`).
- Or document and use a custom code that distinguishes authorization denial from request parsing errors.

Also consider whether the error message should reveal that the tool name is valid-but-blocked ("Tool is not allowed in read-only mode") vs. returning a more opaque denial that does not confirm the tool exists. Current message is convenient for legitimate operators but leaks tool existence to unauthenticated callers.

jwesleye:
Closes #94

## Changes
- Replace JSON-RPC error code `-32600` (Invalid Request) with server-defined `-32001` for tool authorization denial in `http_transport.py`
- Use opaque error message `"Tool call not permitted"` instead of revealing read-only mode details and tool existence
- Add `forbidden_tool: -32001` to error code mapping test
- Update assertion in mutating tools block test to match new code and message

## Test plan
- `uv run pytest tests/http_transport/test_http_server.py tests/http_transport/test_http_mcp_protocol.py -q` — 51 passed
- `uv run ruff check` on all changed files — clean
- `uv run mypy src/open_stocks_mcp/server/http_transport.py` — no issues

### DIFF
--- a/src/open_stocks_mcp/server/http_transport.py
+++ b/src/open_stocks_mcp/server/http_transport.py
@@ -649,11 +649,8 @@ def create_http_server(
                             {
                                 "jsonrpc": "2.0",
                                 "error": {
-                                    "code": -32600,
-                                    "message": (
-                                        "Tool is not allowed in read-only mode. "
-                                        "Enable --allow-trading to permit mutating tools."
-                                    ),
+                                    "code": -32001,
+                                    "message": "Tool call not permitted",
                                 },
                                 "id": request_id,
                             },
@@ -661,7 +658,7 @@ def create_http_server(
                             method=method,
                             request_id=request_id,
                             outcome="error",
-                            error_code=-32600,
+                            error_code=-32001,
                             error_type="forbidden_tool",
                         )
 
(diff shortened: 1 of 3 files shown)


unseriousAI (approved):
FOREMAN_CODE_REVIEW
head_sha: 28409a60f04d6625ea9a42346d34c44e6419e65a
status: PASS
spec_compliance: PASS
blocking_findings: 0
non_blocking_findings: 1

## Summary
The PR correctly implements proper JSON-RPC 2.0 error code semantics for tool authorization denial, moving from `-32600` (Invalid Request) to the implementation-defined `-32001` (Server Error). It also enhances security by using an opaque error message.

## Blocking
None

## Suggestions
1. **Documentation**: Consider documenting implementation-defined JSON-RPC error codes (like `-32001`) in the project's API documentation to assist client developers in handling these specific error states. I've created a follow-up issue for this.

PR #714, opened 2026-05-26
test: add symlink-escape regression test for SchwabBroker._validate_token_path

### LINKED ISSUE #96: Follow-up: add symlink-escape test for SchwabBroker._validate_token_path
Identified during review of PR #88.

**Context**: `_validate_token_path` correctly uses `.resolve()` on both paths before calling `relative_to()`, which handles symlinks under `~/.tokens/` that point outside the allowed root. However, there is no test that exercises this symlink-escape vector.

**Suggested test** (`tests/unit/test_schwab_broker.py`):
```python
def test_rejects_symlink_escaping_tokens_dir(self, tmp_path: Path) -> None:
    """Symlink under ~/.tokens pointing outside is rejected."""
    import os
    tokens_dir = Path.home() / ".tokens"
    tokens_dir.mkdir(parents=True, exist_ok=True)
    link_path = tokens_dir / "escape_link.json"
    target = tmp_path / "outside.json"
    target.write_text("{}")
    try:
        link_path.symlink_to(target)
        with pytest.raises(ValueError, match="token_path must be under"):
            SchwabBroker(api_key="test", app_secret="test", token_path=str(link_path))
    finally:
        link_path.unlink(missing_ok=True)
```

This strengthens security regression coverage for the path-traversal fix.

jwesleye:
Closes #96

## Changes
- Added `test_rejects_symlink_escaping_tokens_dir` to `TestSchwabBroker` in `tests/unit/test_schwab_broker.py`
- Test creates a symlink under `~/.tokens/` pointing to a file outside the tokens directory (`tmp_path/outside.json`)
- Asserts `ValueError` with `"token_path must be under"` is raised when constructing `SchwabBroker` with that symlink path
- Cleans up the symlink in a `finally` block
- Marked `@pytest.mark.unit` and `@pytest.mark.journey_system` for security regression coverage

## Test plan
- `uv run pytest tests/unit/test_schwab_broker.py::TestSchwabBroker::test_rejects_symlink_escaping_tokens_dir -v` — new test passes
- `uv run pytest tests/unit/test_schwab_broker.py -q` — all 29 tests pass, no regressions
- `ruff check` passes with no issues on the modified file

### DIFF
--- a/tests/unit/test_schwab_broker.py
+++ b/tests/unit/test_schwab_broker.py
@@ -189,6 +189,22 @@ class TestSchwabBroker:
                 api_key="test", app_secret="test", token_path="/tmp/outside_token.json"
             )
 
+    @pytest.mark.unit
+    @pytest.mark.journey_system
+    def test_rejects_symlink_escaping_tokens_dir(self, tmp_path: Path) -> None:
+        """Symlink under ~/.tokens pointing outside is rejected."""
+        tokens_dir = Path.home() / ".tokens"
+        tokens_dir.mkdir(parents=True, exist_ok=True)
+        link_path = tokens_dir / "escape_link.json"
+        target = tmp_path / "outside.json"
+        target.write_text("{}")
+        try:
+            link_path.symlink_to(target)
+            with pytest.raises(ValueError, match="token_path must be under"):
+                SchwabBroker(api_key="test", app_secret="test", token_path=str(link_path))
+        finally:
+            link_path.unlink(missing_ok=True)
+
     def test_accepts_token_path_inside_tokens_dir(self) -> None:
         """Test that token_path can be inside ~/.tokens directory."""
         token_dir = Path.home() / ".tokens"
--- a/tests/unit/test_schwab_broker.py
+++ b/tests/unit/test_schwab_broker.py
@@ -193,15 +193,19 @@ class TestSchwabBroker:
     @pytest.mark.journey_system
     def test_rejects_symlink_escaping_tokens_dir(self, tmp_path: Path) -> None:
         """Symlink under ~/.tokens pointing outside is rejected."""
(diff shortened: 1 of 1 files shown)


unseriousAI (dismissed):
FOREMAN_CODE_REVIEW
head_sha: b199f1df4afd825cc07a926b8142c5ef8f0939c4
status: BLOCK
spec_compliance: PASS
blocking_findings: 1
non_blocking_findings: 0

## Summary
Security regression intent is correct, but the test can delete an existing file in the user's real ~/.tokens directory.

## Blocking
1. tests/unit/test_schwab_broker.py:196-206 - The test uses a fixed path under the real Path.home() / ".tokens" and unlinks it in finally. If ~/.tokens/escape_link.json already exists, link_path.symlink_to(target) raises FileExistsError, then the finally block still runs and deletes the pre-existing file or symlink. Token tests should not mutate or clean up arbitrary user-home paths. Patch pathlib.Path.home to tmp_path, matching tests/unit/test_security_umask.py, or otherwise keep the token root entirely inside tmp_path before creating and unlinking the symlink.

## Suggestions
None

unseriousAI:
Addressed blocker: updated symlink-escape test to patch pathlib.Path.home() to a tmp_path-backed fake home before creating/unlinking the symlink, so the test no longer mutates real ~/.tokens paths.\n\nValidation evidence:\n- uv run pytest tests/unit/test_schwab_broker.py::TestSchwabBroker::test_rejects_symlink_escaping_tokens_dir -q --tb=line (exit 0, 1 passed)\n- uv run pytest tests/unit/test_schwab_broker.py -q --tb=line (exit 0, 29 passed)\n\nHead SHA: e325c2f129d6113eed42b67e5be59d6a072f5109

PR #737, opened 2026-05-27
fix: export registry initialization error

### LINKED ISSUE #132: Follow-up: export RegistryNotInitializedError from brokers/__init__.py
Identified during review of PR #111.

`RegistryNotInitializedError` was introduced in PR #111 but was not added to `src/open_stocks_mcp/brokers/__init__.py`'s `__all__`. Callers that need to catch it (e.g. `app.py` after the fix for the broken handler) must import from the internal `open_stocks_mcp.brokers.registry` path instead of the stable package surface.

**Suggested fix**: add `RegistryNotInitializedError` to the `__all__` list in `src/open_stocks_mcp/brokers/__init__.py` and add the import there.

unseriousAI:
Closes #132

## Changes
- Re-export RegistryNotInitializedError from open_stocks_mcp.brokers.
- Add a regression test proving the package export is the same class and appears in __all__.

## Test plan
- uv run pytest tests/unit/test_broker_registry.py -k package_reexports_registry_not_initialized_error -q
- uv run pytest tests/unit/test_broker_registry.py -q
- uv run ruff check src/open_stocks_mcp/brokers/__init__.py tests/unit/test_broker_registry.py
- uv run mypy src/open_stocks_mcp/brokers/__init__.py src/open_stocks_mcp/brokers/registry.py

### DIFF
--- a/src/open_stocks_mcp/brokers/__init__.py
+++ b/src/open_stocks_mcp/brokers/__init__.py
@@ -1,7 +1,11 @@
 """Broker abstraction layer for multi-broker MCP support."""
 
 from open_stocks_mcp.brokers.base import BaseBroker, BrokerAuthStatus
-from open_stocks_mcp.brokers.registry import BrokerRegistry, get_broker_registry
+from open_stocks_mcp.brokers.registry import (
+    BrokerRegistry,
+    RegistryNotInitializedError,
+    get_broker_registry,
+)
 from open_stocks_mcp.brokers.robinhood import RobinhoodBroker
 from open_stocks_mcp.brokers.schwab import SchwabBroker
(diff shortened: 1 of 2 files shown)


jwesleye (approved):
AGENTSHORE_CODE_REVIEW
head_sha: b562dcb13a6174be0b1c28fc1c346687ab9d17c1
status: PASS
spec_compliance: PASS
blocking_findings: 0
non_blocking_findings: 0

## Summary

Minimal, correct fix: re-exports `RegistryNotInitializedError` from the broker package public API with a regression test that verifies identity (not just equality), proving consumers can catch it via the package-level name.

## Blocking

None

## Suggestions

None

PR #738, opened 2026-05-27
feat: expose session pickle-clear circuit breaker state

### LINKED ISSUE #141: Follow-up: expose pickle-clear failure state in get_session_info()
Identified during review of PR #78.

After PR #78 lands, the auth-retry circuit breaker state is invisible to operators. When the circuit breaker trips, the only signal is a CRITICAL log entry. The session_manager.get_session_info() response (surfaced via health/session MCP tools) should include the circuit-breaker fields so operators can observe the blocked state without grepping logs.

Suggested additions to the returned dict:
- 'consecutive_pickle_clear_failures': self._consecutive_pickle_clear_failures
- 'auth_retries_blocked': self.should_block_auth_retries()

unseriousAI:
Closes #141

## Changes
- Added `consecutive_pickle_clear_failures` and `auth_retries_blocked` to `SessionManager.get_session_info()` output.
- Updated `session_status` unit test expectations to assert both fields are surfaced.
- Added a unit test proving `get_session_info()` reflects blocked auth-retry state when pickle-clear failures reach threshold.

## Test plan
- uv run pytest tests/unit/test_stock_market_tools.py -k session_status -v
- uv run pytest tests/unit/test_session_security.py -k get_session_info_includes_pickle_clear_circuit_breaker_state -v
- uv run pytest tests/unit/ -q
- uv run ruff check src/open_stocks_mcp/brokers/session_state.py
- uv run mypy src/open_stocks_mcp/brokers/session_state.py

### DIFF
--- a/src/open_stocks_mcp/brokers/session_state.py
+++ b/src/open_stocks_mcp/brokers/session_state.py
@@ -203,6 +203,8 @@ class SessionManager:
             "session_timeout_hours": self.session_timeout_hours,
             "failed_login_attempts": self._failed_login_attempts,
             "max_failed_attempts": self.max_failed_attempts,
+            "consecutive_pickle_clear_failures": self._pickle._consecutive_pickle_clear_failures,
+            "auth_retries_blocked": self.should_block_auth_retries(),
         }
 
         if self.login_time:
--- a/tests/unit/test_session_security.py
+++ b/tests/unit/test_session_security.py
@@ -241,3 +241,19 @@ class TestClearPickleFile:
         manager.reset_clear_failures()
 
         assert manager.should_block_auth_retries() is False
+
+
+class TestSessionInfoCircuitBreaker:
+    def test_get_session_info_includes_pickle_clear_circuit_breaker_state(
+        self, manager: SessionPickleManager
+    ) -> None:
+        session_manager = SessionManager(pickle_manager=manager)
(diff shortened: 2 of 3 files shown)


jwesleye (approved):
AGENTSHORE_CODE_REVIEW
head_sha: ea6a85738d6b0d9757441ec816119bbba1c8b135
status: PASS
spec_compliance: PASS
blocking_findings: 0
non_blocking_findings: 2

## Summary
Clean, minimal change that correctly surfaces circuit-breaker state through `get_session_info()`. Tests pass, mypy clean, ruff clean.

## Blocking
None

## Suggestions

1. **`session_state.py:206` — private attribute access**: `self._pickle._consecutive_pickle_clear_failures` reaches into a private attribute. `SessionPickleManager` already exposes the boolean predicate via the public `should_block_auth_retries()`; adding a companion `get_consecutive_clear_failures() -> int` accessor would keep the boundary consistent and insulate `SessionManager` from future renames inside `SessionPickleManager`.

2. **`test_session_security.py:255-258` — indirect assertion**: The test sets `manager._consecutive_pickle_clear_failures = manager._max_pickle_clear_failures` and then asserts `info["consecutive_pickle_clear_failures"] == manager._consecutive_pickle_clear_failures`. Since both sides reference the same variable, the assertion is tautological about the specific value. Asserting `== manager._max_pickle_clear_failures` directly (or the literal `3`) would better communicate intent and catch a regression where the field is wired to the wrong counter.

PR #741, opened 2026-05-27
feat: make pickle-clear failure threshold a constructor parameter

### LINKED ISSUE #140: Follow-up: make pickle-clear failure threshold a constructor parameter
Identified during review of PR #78.

The  threshold is hardcoded to  in . The existing  is already a constructor parameter — for consistency, testability, and caller flexibility the pickle-clear threshold should follow the same pattern.

**Suggested change:**
```python
def __init__(self, session_timeout_hours: int = 23, max_failed_attempts: int = 3, max_pickle_clear_failures: int = 3):
    ...
    self._max_pickle_clear_failures = max_pickle_clear_failures
```

This also makes unit tests for the circuit breaker simpler — tests can pass  rather than simulating 3 failures.

unseriousAI:
Closes #140

## Changes
- Updated `SessionPickleManager` to accept `max_pickle_clear_failures` in its constructor.
- Updated `SessionManager` to accept `max_pickle_clear_failures` and pass it down.
- Added test coverage to verify that the threshold is configurable and defaults to 3.

## Test plan
- `uv run pytest tests/unit/test_session_manager.py -k 'pickle_clear_failure_threshold'`
- `uv run ruff check` and `uv run mypy`

### DIFF
--- a/src/open_stocks_mcp/brokers/session_pickle.py
+++ b/src/open_stocks_mcp/brokers/session_pickle.py
@@ -12,9 +12,9 @@ from open_stocks_mcp.logging_config import logger
 class SessionPickleManager:
     """Manages pickle read/write/clear and Fernet encryption for session tokens."""
 
-    def __init__(self) -> None:
+    def __init__(self, max_pickle_clear_failures: int = 3) -> None:
         self._consecutive_pickle_clear_failures = 0
-        self._max_pickle_clear_failures = 3
+        self._max_pickle_clear_failures = max_pickle_clear_failures
 
     def _get_tokens_dir(self) -> Path:
         tokens_dir = Path.home() / ".tokens"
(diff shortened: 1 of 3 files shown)


jwesleye (approved):
AGENTSHORE_CODE_REVIEW
head_sha: 7c75a12909d8b5554ec3b81c5616676244d5b3c9
status: PASS
spec_compliance: PASS
blocking_findings: 0
non_blocking_findings: 2

## Summary
Clean, minimal change that makes the pickle-clear failure threshold configurable — exactly what issue #140 requested. Tests pass, linting clean, mypy clean. Approved.

## Blocking
None

## Suggestions
1. **`tests/unit/test_session_manager.py:469-470, 480-482`** — Missing blank line between top-level test functions. PEP 8 (and the existing file convention) uses two blank lines between module-level function definitions. The two new tests each have only one blank line separating them from the preceding function.

2. **`src/open_stocks_mcp/brokers/session_state.py:30`** — `self._max_pickle_clear_failures` is stored on `SessionManager` but only used to initialize the fallback `SessionPickleManager`. When a `pickle_manager` is injected via DI, this attribute becomes dead state. Consider either not storing it as an instance attribute (just use a local variable), or documenting that it's ignored when `pickle_manager` is provided. Low priority — the current code is correct, just slightly misleading.

PR #746, opened 2026-05-27
fix: replace assert statements with explicit guards in schwab_stream.py

### LINKED ISSUE #202: Follow-up: replace assert statements with explicit guards in schwab_stream.py
Identified during review of PR #186.

`assert self.stream_client is not None` appears on lines 75 and 127 of `src/open_stocks_mcp/brokers/schwab_stream.py`. Both occur after code paths that should guarantee non-None, but `assert` is stripped when Python is run with `-O`, removing the safety check in optimized deployments.

Replace each with an explicit guard:
```python
if self.stream_client is None:
    return  # or raise RuntimeError("stream_client unexpectedly None")
```

jwesleye:
Closes #202

## Changes
- Replaced `assert self.stream_client is not None` on line 77 of `schwab_stream.py` with an explicit `if self.stream_client is None:` guard that logs a clear error, sets `_is_running = False`, and returns `False` before any handler registration — ensuring correct behavior under `python -O`.
- Replaced `assert self.stream_client is not None` on line 133 of `schwab_stream.py` with an explicit `if self.stream_client is None:` guard that logs a clear error and `break`s the reconnect loop — preventing a silent `AttributeError` under `python -O`.
- Added `test_schwab_stream_manager_start_handles_missing_stream_client` — verifies that when `StreamClient()` returns `None`, `start()` returns `False`, `is_running` stays `False`, and no background task is created.
- Added `test_schwab_stream_manager_run_loop_skips_relogin_when_stream_client_disappears` — verifies that when `stream_client` is cleared mid-loop, `_run_loop()` exits cleanly without touching the None client.

## Test plan
- `uv run pytest tests/unit/test_capabilities_and_streaming.py -k "schwab_stream_manager_start_handles_missing_stream_client or schwab_stream_manager_run_loop_skips_relogin_when_stream_client_disappears" -q` — 2 passed
- `uv run pytest tests/unit/test_capabilities_and_streaming.py -k schwab_stream_manager -q` — 13 passed (all existing tests intact)
- `uv run ruff check src/open_stocks_mcp/brokers/schwab_stream.py tests/unit/test_capabilities_and_streaming.py` — no issues
- `uv run mypy src/open_stocks_mcp/brokers/schwab_stream.py tests/unit/test_capabilities_and_streaming.py` — no issues

### DIFF
--- a/src/open_stocks_mcp/brokers/schwab_stream.py
+++ b/src/open_stocks_mcp/brokers/schwab_stream.py
@@ -74,7 +74,13 @@ class SchwabStreamManager:
                         logger.error(f"Error in Schwab stream custom handler: {e}")
 
             # Register handlers for Level One, Level 2 and account activity
-            assert self.stream_client is not None
+            if self.stream_client is None:
+                logger.error(
+                    "Schwab StreamClient unexpectedly None after construction;"
+                    " cannot start streamer"
+                )
+                self._is_running = False
+                return False
             self.stream_client.add_level_one_equity_handler(_core_handler)
             self.stream_client.add_level_one_option_handler(_core_handler)
             self.stream_client.add_nasdaq_book_handler(_core_handler)
@@ -129,8 +135,13 @@ class SchwabStreamManager:
                 await asyncio.sleep(5)  # Wait before retry
                 if self._is_running:
                     # Attempt re-login
+                    if self.stream_client is None:
+                        logger.error(
+                            "Schwab stream_client is None before re-login;"
+                            " exiting reconnect loop"
+                        )
+                        break
                     try:
-                        assert self.stream_client is not None
                         await self.stream_client.login()
                     except Exception as re_login_err:
                         logger.error(
--- a/src/open_stocks_mcp/brokers/schwab_stream.py
+++ b/src/open_stocks_mcp/brokers/schwab_stream.py
@@ -140,6 +140,7 @@ class SchwabStreamManager:
                             "Schwab stream_client is None before re-login;"
                             " exiting reconnect loop"
                         )
+                        self._is_running = False
                         break
                     try:
                         await self.stream_client.login()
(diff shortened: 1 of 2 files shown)


unseriousAI (dismissed):
AGENTSHORE_CODE_REVIEW
head_sha: 2df3c7c1f1ee4532d8c5aab24aea72187bdbbf3f
status: BLOCK
spec_compliance: PASS
blocking_findings: 1
non_blocking_findings: 0

## Summary
The assert replacements are present, but the reconnect-loop guard can leave the manager reporting it is still running after the background task exits.

## Blocking
1. src/open_stocks_mcp/brokers/schwab_stream.py:138-143 - When `stream_client` disappears after `handle_responses()` fails, this guard logs and breaks out of `_run_loop()` without clearing `_is_running`. That leaves `manager.is_running == True` even though the stream task has exited and `stream_client` is `None`; a later `start()` call will return `True` immediately at the early `if self._is_running` check instead of recreating the stream client. Set `_is_running = False` before exiting this path, and update the regression test so `asyncio.sleep` does not clear `_is_running` before the new guard runs; the test should assert the manager is no longer running after this failure.

## Suggestions
None

jwesleye:
Addressed blocking finding from AgentShore review:

**Fix**: Added `self._is_running = False` before `break` at the `stream_client is None` guard in `_run_loop()` (schwab_stream.py:138-144). Without this, a subsequent `start()` call would return `True` immediately at the early `_is_running` check instead of recreating the stream client.

**Test update**: Replaced `fast_sleep` (which was itself clearing `_is_running`, masking the guard) with a no-op sleep mock so the guard is now the sole mechanism that clears `_is_running`. Added `assert manager.is_running is False` to verify the state after the failure path.

**Validation**:
- `uv run pytest tests/unit/test_capabilities_and_streaming.py -k schwab_stream_manager -q` → 13 passed
- `uv run ruff check src/open_stocks_mcp/brokers/schwab_stream.py tests/unit/test_capabilities_and_streaming.py` → no issues
- `uv run mypy src/open_stocks_mcp/brokers/schwab_stream.py tests/unit/test_capabilities_and_streaming.py` → no issues

Head: 72addb5ab7ec61e46ba957eace48e3407b61a56e

PR #819, opened 2026-05-27
test: strengthen SSE endpoint wiring assertions

### LINKED ISSUE #259: Follow-up: strengthen SSE endpoint test beyond route registration

`tests/http_transport/test_http_server.py:138-142` previously exercised `GET /sse` via the HTTP client; PR #239 replaced that with a route-table check (`assert any(getattr(route, 'path', None) == '/sse' ...)`). The change was made because the original assertion was hanging, which is reasonable, but the new test will pass even if the SSE handler is missing or crashes immediately.

Suggested fix: add a lighter behavior check — e.g., verify the route's `endpoint` is callable / matches the expected handler reference, or use a short read timeout against `GET /sse` so handler-wiring regressions still surface without re-introducing the original hang.

unseriousAI:
Closes #259

## Changes
- Replace the `/sse` path-only assertion with route lookup and method checks
- Assert the `/sse` endpoint is callable and bound to `sse_endpoint`
- Invoke the route endpoint directly with a minimal Starlette request
- Assert `EventSourceResponse` shape and validate the first `connected` event payload
- Close the async body iterator after one event to avoid hangs

## Test plan
- uv run pytest tests/http_transport/test_http_server.py::TestMCPIntegration::test_sse_endpoint -q
- uv run pytest tests/http_transport/test_http_server.py::TestMCPIntegration::test_sse_endpoint tests/http_transport/test_http_integration.py::TestHTTPIntegration::test_mcp_endpoint_accessibility -q
- uv run ruff check tests/http_transport/test_http_server.py

### DIFF
--- a/tests/http_transport/test_http_server.py
+++ b/tests/http_transport/test_http_server.py
@@ -9,6 +9,8 @@ from unittest.mock import AsyncMock, Mock, patch
 import httpx
 import pytest
 from mcp.server.fastmcp import FastMCP
+from sse_starlette import EventSourceResponse
+from starlette.requests import Request
 
 import open_stocks_mcp.tracing as tracing_module
 from open_stocks_mcp import __version__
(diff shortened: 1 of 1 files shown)


jwesleye (approved):
AGENTSHORE_CODE_REVIEW
head_sha: f1be2154dcd6a115d03328525d162b3212fdab83
status: PASS
spec_compliance: PASS
blocking_findings: 0
non_blocking_findings: 0

## Summary

Clean, well-scoped test improvement that addresses every requirement from #259. The test now verifies route method, handler reference, endpoint callability, `EventSourceResponse` type and media type, and first `connected` event payload — all without re-introducing the original SSE hang. Iterator cleanup via `try/finally` with `contextlib.suppress` is correct. All 34 tests in the file pass, ruff is clean.

## Blocking

None.

## Suggestions

None.

PR #864, opened 2026-05-27
fix: drop redundant auth retry override

### LINKED ISSUE #257: Follow-up: clarify or drop redundant auth retry config field

`RetryConfig` in `src/open_stocks_mcp/config.py:44-45` exposes two auth-related knobs:
- `auth_max_retries: int = 1` — used directly in `execute_with_retry` as `max_auth_retries` for the re-authentication counter.
- `authentication_max_retries: int | None = None` — feeds `RetryConfig.max_retries_for("authentication")`, but the auth branch in `tools/retry.py` returns/raises before the per-error override is consulted. The override only takes effect when `handle_auth_errors=False`.

Result: setting `OPEN_STOCKS_MCP_RETRY_AUTHENTICATION_MAX_RETRIES` does nothing in the default path, and two similarly-named env vars are easy to confuse.

Suggested fix: either drop `authentication_max_retries` (and the corresponding env var), or rename `auth_max_retries` for clarity and document which knob applies when.

unseriousAI:
Closes #257

## Changes
- Removed the dead `RetryConfig.authentication_max_retries` field and authentication-specific retry override.
- Stopped reading `OPEN_STOCKS_MCP_RETRY_AUTHENTICATION_MAX_RETRIES` in `load_config()`.
- Added a regression test covering field removal and authentication fallback to generic `max_retries`.

## Test plan
- `uv run pytest tests/unit/test_config.py -q --tb=short`
- `uv run ruff check src/open_stocks_mcp/config.py tests/unit/test_config.py`
- `uv run mypy src/open_stocks_mcp/config.py --show-error-codes --no-error-summary`
- `uv run pytest tests/unit/test_error_handling.py -q --tb=short`

## Notes
- `uv run pytest tests/unit/ -q --tb=short` was run and currently fails in unrelated cache, dependency pin, missing `SCHWAB_STATUS.md`, and docs reference checks; this PR only changes retry config and config tests.

### DIFF
--- a/src/open_stocks_mcp/config.py
+++ b/src/open_stocks_mcp/config.py
@@ -181,7 +181,6 @@ class RetryConfig:
     initial_delay: float = 1.0
     backoff_factor: float = 2.0
     auth_max_retries: int = 1
-    authentication_max_retries: int | None = None
     network_max_retries: int | None = None
     rate_limit_max_retries: int | None = None
     api_max_retries: int | None = None
@@ -189,7 +188,6 @@ class RetryConfig:
 
     def max_retries_for(self, error_type: str) -> int:
         overrides = {
-            "authentication": self.authentication_max_retries,
             "network": self.network_max_retries,
             "rate_limit": self.rate_limit_max_retries,
             "api": self.api_max_retries,
@@ -499,9 +497,6 @@ def load_config(config_path: Path | str | None = None) -> ServerConfig:
             auth_max_retries=_parse_int(
                 os.getenv("OPEN_STOCKS_MCP_RETRY_AUTH_MAX_RETRIES", "1"),
                 "OPEN_STOCKS_MCP_RETRY_AUTH_MAX_RETRIES",
-            ),
-            authentication_max_retries=_env_optional_int(
-                "OPEN_STOCKS_MCP_RETRY_AUTHENTICATION_MAX_RETRIES"
             ),
             network_max_retries=_env_optional_int(
                 "OPEN_STOCKS_MCP_RETRY_NETWORK_MAX_RETRIES"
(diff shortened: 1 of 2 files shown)


jwesleye (approved):
AGENTSHORE_CODE_REVIEW
head_sha: 1fafe816f817d896f999a7135230de3d75829e0f
status: PASS
spec_compliance: PASS
blocking_findings: 0
non_blocking_findings: 1

## Summary
Clean, correct removal of a dead config field — spec-compliant and well-tested.

## Blocking
None

## Suggestions
1. **tests/unit/test_config.py:179** — `Path("/tmp/definitely-missing-config.yaml")` is a hardcoded sentinel. On a system where that path accidentally exists the test would silently load real config and the `max_retries == 3` assertions could mismatch. Prefer `tmp_path / "nonexistent.yaml"` (already a fixture in sibling tests) for hermetic isolation. Non-blocking because the current name is highly unlikely to collide.

jwesleye (approved):
AGENTSHORE_CODE_REVIEW
head_sha: 1fafe816f817d896f999a7135230de3d75829e0f
status: PASS
spec_compliance: PASS
blocking_findings: 0
non_blocking_findings: 1

## Summary
Clean, correct removal of a dead config field — spec-compliant and well-tested.

## Blocking
None

## Suggestions
1. **tests/unit/test_config.py:179** — `Path("/tmp/definitely-missing-config.yaml")` is a hardcoded sentinel. On a system where that path accidentally exists the test would silently load real config and the assertions could mismatch. Prefer `tmp_path / "nonexistent.yaml"` for hermetic isolation. Non-blocking because the current name is highly unlikely to collide.

PR #877, opened 2026-05-27
perf: cache retry config in execute_with_retry (#258)

### LINKED ISSUE #258: Follow-up: cache load_config() to avoid re-reading env on every retry call

`execute_with_retry` in `src/open_stocks_mcp/tools/retry.py:34` calls `load_config()` on every invocation, which re-reads up to 9 `OPEN_STOCKS_MCP_RETRY_*` environment variables each time. Per-call overhead is small but unnecessary — config is effectively immutable per-process.

Suggested fix: wrap `load_config()` with `functools.lru_cache(maxsize=1)` (it has no args), or compute the `RetryConfig` once at module load. Either preserves the env-var-driven defaults while removing the per-call work.

jwesleye:
Closes #258

## Changes
- Added `_get_retry_config()` helper in `src/open_stocks_mcp/tools/retry.py` decorated with `@functools.lru_cache(maxsize=1)` that wraps `load_config().retry` and raises `RuntimeError` if the retry section is absent
- Replaced the inline `load_config().retry` call and its `None` guard in `execute_with_retry` with a call to `_get_retry_config()`, so retry defaults are read from env vars once per process instead of on every invocation
- Updated `_patch_retry_dependencies` in `tests/unit/test_error_handling.py` to call `_get_retry_config.cache_clear()` so env-mutating tests pick up fresh config
- Added `test_execute_with_retry_caches_retry_config` proving that two back-to-back `execute_with_retry` calls result in exactly one `load_config` invocation

## Test plan
- `uv run pytest tests/unit/test_error_handling.py -k "configured_retry_defaults or per_error_class_retry_override or explicit_arguments_override_config or caches_retry_config" -q --tb=line` — 4 passed
- `uv run pytest tests/unit/test_error_handling.py tests/unit/test_circuit_breaker.py -q --tb=line` — 111 passed
- `uv run ruff check src/open_stocks_mcp/tools/retry.py tests/unit/test_error_handling.py` — clean
- `uv run mypy src/open_stocks_mcp/tools/retry.py` — no issues

### DIFF
--- a/src/open_stocks_mcp/tools/retry.py
+++ b/src/open_stocks_mcp/tools/retry.py
@@ -5,7 +5,7 @@ import functools
 from collections.abc import Callable
 from typing import Any
 
-from open_stocks_mcp.config import load_config
+from open_stocks_mcp.config import RetryConfig, load_config
 from open_stocks_mcp.logging_config import logger
 from open_stocks_mcp.tools.exceptions import (
     AuthenticationError,
@@ -15,6 +15,14 @@ from open_stocks_mcp.tools.exceptions import (
 )
 
 DEFAULT_MAX_RETRIES = 3
+
+
+@functools.lru_cache(maxsize=1)
+def _get_retry_config() -> RetryConfig:
+    config = load_config().retry
+    if config is None:
+        raise RuntimeError("Retry configuration unavailable")
+    return config
 
 
 async def execute_with_retry(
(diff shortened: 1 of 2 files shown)


unseriousAI (approved):
AGENTSHORE_CODE_REVIEW
head_sha: 0a20b0ccd2f5561547e5a3cb820d5d6736461004
status: PASS
spec_compliance: PASS
blocking_findings: 0
non_blocking_findings: 0

## Summary
Approve: change matches issue #258 and safely caches retry config to avoid repeated env/config reads.

## Blocking
None

## Suggestions
None

PR #878, opened 2026-05-27
fix: add integration marker to multi-broker tests

### LINKED ISSUE #274: QA: 8 multi-broker integration tests fail - test expectations diverged from registry/broker API
## Failure

All 8 tests in `tests/integration/test_multi_broker.py::TestMultiBrokerIntegration` fail because the test expectations no longer match the current `BrokerRegistry`, `RobinhoodBroker`, `SchwabBroker`, and `BrokerAuthStatus` API.

Secondary marker-hygiene issue: the file has no `@pytest.mark.integration` decorator, so these tests run even with `pytest -m 'not integration'` (verified — `grep -l 'pytest.mark.integration' tests/integration/*.py` only returns `tests/integration/test_basic_api.py`).

## Evidence

Failures (from `uv run pytest -m "not slow and not exception_test and not integration and not live_market and not login_flow and not agent_evaluation" --tb=line -v --timeout=15 -o "addopts="`):

- `test_authenticate_all_brokers` — `AttributeError: module 'open_stocks_mcp.brokers.schwab' does not have the attribute 'auth'` (test patches `brokers.schwab.auth` which isn't a top-level attribute of that module).
- `test_partial_authentication_failure` — same `brokers.schwab.auth` AttributeError.
- `test_broker_not_found_error` — `Failed: DID NOT RAISE <class 'ValueError'>`. Test expects `registry.get_broker("nonexistent")` to raise `ValueError(match="Broker not found")`, but `BrokerRegistry.get_broker` (src/open_stocks_mcp/brokers/registry.py:42-61) returns `None` and only logs a warning.
- `test_set_active_broker_not_registered` — `Failed: DID NOT RAISE <class 'ValueError'>`. Test expects `registry.set_active_broker("schwab")` to raise `ValueError(match="Broker not registered")`, but `BrokerRegistry.set_active_broker` (registry.py:90-105) returns `False` and logs an error.
- `test_broker_auth_status_tracking` — `AttributeError: 'RobinhoodBroker' object has no attribute 'auth_status'`. The base class exposes `auth_info.status`, not `auth_status`. Test also references `BrokerAuthStatus.UNAUTHENTICATED`, but the enum defines `NOT_AUTHENTICATED` (src/open_stocks_mcp/brokers/base.py:14).
- `test_concurrent_broker_operations` — same `brokers.schwab.auth` AttributeError.
- `test_global_registry_singleton` — `assert <coroutine object get_broker_registry at 0x...> is <coroutine object get_broker_registry at 0x...>`. Test treats `get_broker_registry()` as sync, but it is `async def` (src/open_stocks_mcp/brokers/registry.py:273).
- `test_schwab_token_persistence` — same `brokers.schwab.auth` AttributeError.

## Reproduction

```
uv sync
uv pip install pytest-timeout
uv run pytest tests/integration/test_multi_broker.py -v --tb=short --timeout=15 -o "addopts="
```

## Root cause grouping

1. **Registry API drift (4 tests)** — `get_broker`/`set_active_broker` return values vs. raising `ValueError`, and sync vs. async `get_broker_registry`. Either tests need to be updated to match the current return-value contract, or the registry should raise (see also #209 graceful-auth design which keeps the return-value style).
2. **Broker attribute drift (1 test)** — `broker.auth_status` vs. `broker.auth_info.status`, and `BrokerAuthStatus.UNAUTHENTICATED` vs. `NOT_AUTHENTICATED`.
3. **Schwab module import path (4 tests)** — `patch("open_stocks_mcp.brokers.schwab.auth")` doesn't resolve. `brokers/schwab.py` imports `schwab.auth` under a different name; tests should patch the real import path (likely `open_stocks_mcp.brokers.schwab.schwab_auth` or whatever the local binding is — verify in `brokers/schwab.py`).
4. **Marker hygiene** — file is in `tests/integration/` but lacks `@pytest.mark.integration`, so it runs under default `pytest`. Either add the marker class-wide, or move the file out of `tests/integration/` if it's truly mock-only.

## Branch

`main` @ `bc6526b`

jwesleye:
Closes #274

## Changes
- Added `pytestmark = pytest.mark.integration` to `tests/integration/test_multi_broker.py` so that the 26 multi-broker integration tests are properly excluded by marker-based filters (`-m "not integration"`)

## Context
The 8 test failures described in #274 (registry API drift, broker attribute drift, Schwab module import path) have already been fixed on `main`. The remaining issue was marker hygiene: the file lacked `@pytest.mark.integration`, causing these integration tests to run even when explicitly excluded.

## Test plan
- `uv run pytest tests/integration/test_multi_broker.py -v --timeout=15 -o "addopts="` — all 26 tests pass
- `uv run pytest tests/integration/test_multi_broker.py --collect-only -q -m "not integration" -o "addopts="` — 26 deselected, none collected
- `uv run pytest tests/unit/test_broker_registry.py tests/unit/test_broker_base.py tests/unit/test_schwab_broker.py -q -o "addopts="` — 79 related tests pass
- `uv run ruff check tests/integration/test_multi_broker.py` — clean

### DIFF
--- a/tests/integration/test_multi_broker.py
+++ b/tests/integration/test_multi_broker.py
@@ -21,6 +21,8 @@ from open_stocks_mcp.server.tool_helpers import (
 )
 from tests.conftest import MockBroker
 
+pytestmark = pytest.mark.integration
+
 
 class TestMultiBrokerIntegration:
     """Test multi-broker scenarios."""


unseriousAI (approved):
AGENTSHORE_CODE_REVIEW
head_sha: 0758b076ae6dfb2fea2c7a171a985da62c3e74ec
status: PASS
spec_compliance: PASS
blocking_findings: 0
non_blocking_findings: 0

## Summary
Marker-only change correctly scopes multi-broker integration tests to integration runs.

## Blocking
None

## Suggestions
None

PR #882, opened 2026-05-27
test: add rate limiter isolation regression coverage

### LINKED ISSUE #275: QA: 4 option-position unit tests time out via shared rate limiter (no isolation between tests)
## Failure

Four `tests/unit/test_options_tools.py::TestOptionPositions` tests fail with `pytest-timeout (>15.0s)` because the global `get_rate_limiter()` instance accumulates call history across the test session. By the time these tests run, the per-minute bucket is full and `execute_with_retry` blocks for ~30-50s waiting for the bucket to drain.

## Evidence

Captured stderr from `test_get_aggregate_positions_empty`:

```
INFO     open_stocks_mcp:robinhood_options_tools.py:539 Getting aggregated option positions
WARNING  open_stocks_mcp:rate_limiter.py:79 Rate limit reached (minute). Waiting 48.7s
Failed: Timeout (>15.0s) from pytest-timeout.
```

`test_get_all_option_positions_success` shows "Waiting 33.7s"; `test_get_open_option_positions_success` "Waiting >15s"; `test_get_open_option_positions_none` same.

All four failures cluster around the rate limiter waiting on a shared singleton (`get_rate_limiter()` from `src/open_stocks_mcp/tools/rate_limiter.py`). `execute_with_retry` calls `rate_limiter.acquire(endpoint)` unconditionally (src/open_stocks_mcp/tools/error_handling.py:230-231), and these tests don't patch out the rate limiter the way other passing tests do.

The passing tests in the same file (`test_get_open_option_positions_with_details_success`, `test_get_open_option_positions_with_details_no_positions`) clearly mock or short-circuit the rate-limited path. The 4 failing tests do not, so they're vulnerable to test-ordering effects.

Failing tests:
- `tests/unit/test_options_tools.py::TestOptionPositions::test_get_aggregate_positions_empty`
- `tests/unit/test_options_tools.py::TestOptionPositions::test_get_all_option_positions_success`
- `tests/unit/test_options_tools.py::TestOptionPositions::test_get_open_option_positions_success`
- `tests/unit/test_options_tools.py::TestOptionPositions::test_get_open_option_positions_none`

## Reproduction

```
uv sync
uv pip install pytest-timeout
uv run pytest -m "not slow and not exception_test and not integration and not live_market and not login_flow and not agent_evaluation" --tb=line -v --timeout=15 -o "addopts="
```

## Root cause

The rate limiter is a module-level singleton with no test fixture that resets it. As tests run, the rate-limiter bucket fills up; later tests that don't mock `acquire` block on the real `asyncio.sleep(...)`.

## Suggested fix

1. Add an autouse fixture (in `tests/conftest.py` or `tests/unit/conftest.py`) that resets the rate-limiter singleton or replaces it with a no-op for unit tests.
2. Alternatively, have these four tests mock `open_stocks_mcp.tools.error_handling.get_rate_limiter` or `open_stocks_mcp.tools.rate_limiter.get_rate_limiter` the same way the passing options tests do.

Related: #233 (pytest-timeout), #167 (circuit breaker), but neither covers the test-isolation root cause.

## Branch

`main` @ `bc6526b`

unseriousAI:
Closes #275

## Changes
- add `tests/unit/test_rate_limiter_isolation.py` with journey-marked regression coverage
- verify `RateLimiter.reset()` clears minute/hour usage and endpoint buckets
- add sequential tests proving autouse `reset_tool_state` prevents cross-test limiter bleed
- add async performance regression test that exercises previously flaky option-position paths and asserts each returns in under 2 seconds

## Test plan
- `uv run pytest tests/unit/test_rate_limiter_isolation.py -v`
- `uv run ruff check tests/unit/test_rate_limiter_isolation.py`
- `uv run mypy tests/unit/test_rate_limiter_isolation.py`
- `uv run pytest tests/unit/test_options_tools.py::TestOptionPositions -v`
- `uv run pytest -m "not slow and not exception_test" --tb=short -q` *(currently fails on pre-existing cache/dependency/docs assertions unrelated to #275)*

### DIFF
--- a/tests/unit/test_rate_limiter_isolation.py
+++ b/tests/unit/test_rate_limiter_isolation.py
@@ -1,0 +1,123 @@
+"""Regression tests for rate-limiter isolation across unit tests."""
+
+import asyncio
+import time
+from collections.abc import Awaitable, Callable
+from time import perf_counter
+from typing import Any
+from unittest.mock import patch
+
+import pytest
+
+from open_stocks_mcp.tools.options.positions import (
+    get_aggregate_positions,
+    get_all_option_positions,
+    get_open_option_positions,
+)
+from open_stocks_mcp.tools.rate_limiter import get_rate_limiter
+
+_bucket_filled_in_previous_test = False
+
+
+@pytest.mark.journey_system
+@pytest.mark.unit
+def test_rate_limiter_reset_clears_state() -> None:
+    """Reset clears all global limiter state."""
+    limiter = get_rate_limiter()
+    now = time.time()
+
+    for _ in range(limiter.calls_per_minute):
+        limiter.call_times.append(now)
+    limiter.endpoint_buckets["options"].append(now)
+
+    assert limiter.get_stats()["calls_last_minute"] == limiter.calls_per_minute
+    assert limiter.endpoint_buckets
+
+    limiter.reset()
+    stats = limiter.get_stats()
+
+    assert stats["calls_last_minute"] == 0
+    assert stats["calls_last_hour"] == 0
+    assert limiter.endpoint_buckets == {}
+
+
+@pytest.mark.journey_system
+@pytest.mark.unit
+def test_autouse_fixture_seed_rate_limiter_for_next_test() -> None:
+    """Seed the limiter so the next test can prove fixture isolation."""
+    global _bucket_filled_in_previous_test
+
+    limiter = get_rate_limiter()
+    now = time.time()
+    for _ in range(limiter.calls_per_minute):
+        limiter.call_times.append(now)
+
+    _bucket_filled_in_previous_test = True
+    assert limiter.get_stats()["calls_last_minute"] == limiter.calls_per_minute
+
+
+@pytest.mark.journey_system
+@pytest.mark.unit
+def test_autouse_fixture_prevents_cross_test_bleed() -> None:
+    """Autouse reset fixture should clear global limiter state between tests."""
+    assert _bucket_filled_in_previous_test is True
+
+    limiter = get_rate_limiter()
+    stats = limiter.get_stats()
+
+    assert stats["calls_last_minute"] == 0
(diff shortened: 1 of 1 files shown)


jwesleye (approved):
AGENTSHORE_CODE_REVIEW
head_sha: 89b67b70b7f610ef209e5274763ff37e2bbaca95
status: PASS
spec_compliance: PASS
blocking_findings: 0
non_blocking_findings: 3

## Summary
All 4 tests pass; ruff and mypy clean; spec requirements for #275 fully met — approving.

## Blocking
None

## Suggestions

1. **`test_rate_limiter_isolation.py:121–122` — tautological `elapsed < 2.0` assertion**
   `asyncio.wait_for(tool_fn(), timeout=2.0)` already raises `TimeoutError` if the call takes ≥ 2 s, so the subsequent `assert elapsed < 2.0` can never be False after a successful `wait_for`. Remove the redundant assertion, or widen it to `assert elapsed < 2.5` as a loose wall-clock safety net that tolerates OS scheduling jitter while still catching pathological slowdowns.

2. **`test_rate_limiter_isolation.py:18–71` — ordering dependency via module-level global**
   `_bucket_filled_in_previous_test` couples `test_autouse_fixture_prevents_cross_test_bleed` to `test_autouse_fixture_seed_rate_limiter_for_next_test` running first. If a randomising plugin (e.g. `pytest-randomly`) is added later the bleed-check test will fail with an unhelpful `AssertionError: assert False is True`. Consider adding `# tests are intentionally ordered — do not reorder` inline or adding a `pytest.mark.order` guard.

3. **`test_rate_limiter_isolation.py:76` — redundant `@pytest.mark.asyncio`**
   `pyproject.toml` already sets `asyncio_mode = "auto"`, which makes `@pytest.mark.asyncio` a no-op on async test functions. Removing it keeps the file consistent with other async tests in the suite.

PR #296, opened 2026-05-20
feat: add per-tool MCP metrics endpoint and instrumentation

unseriousAI:
Closes #162

## Changes
- add per-tool duration tracking in MetricsCollector with per-tool p50/p95/p99 and calls-per-minute
- add Prometheus formatter with stable metric names and escaped label handling
- instrument HTTP JSON-RPC `tools/call` requests to record tool latency/success/failure
- add unauthenticated `GET /metrics` endpoint and expose it in root endpoint metadata/startup logs
- add unit + HTTP tests covering collector metrics and `/metrics` output
- document metrics scrape usage in README

## Test plan
- `uv run pytest tests/unit/test_monitoring.py -q`
- `uv run pytest tests/http_transport/test_http_server.py -q -k 'metrics or tool_call'`
- `uv run ruff check src/open_stocks_mcp/monitoring.py src/open_stocks_mcp/server/http_transport.py tests/unit/test_monitoring.py tests/http_transport/test_http_server.py`
- `uv run ruff format --check src/open_stocks_mcp/monitoring.py src/open_stocks_mcp/server/http_transport.py tests/unit/test_monitoring.py tests/http_transport/test_http_server.py`
- `uv run mypy .`

## Notes
- Running the broader combined suite `uv run pytest tests/unit/test_monitoring.py tests/http_transport/test_http_server.py -q` intermittently hits a pre-existing live-server assertion in `TestLiveHTTPServer.test_server_startup` where `/health` may return `degraded` instead of `healthy`.

PR #888, opened 2026-05-27
docs: document percentile calculation change (#304)

### LINKED ISSUE #304: Follow-up: document silent change to global percentile values in get_metrics()

The new `MetricsCollector._percentile` helper now backs the pre-existing `p50/p95/p99_response_time_ms` keys in `get_metrics()`. The old code used `samples[int(len*N)]` (truncated fraction index); the new helper uses ceil-rank. Values differ for short or even-length sample lists. Any consumer of those keys (status page, dashboards) may see a small drift.

Action: add a note to the release/changelog when this PR ships, or, if the prior behavior is preferred for those legacy keys, restore the original integer-index lookup there and keep `_percentile` only for the new per-tool fields.

jwesleye:
Closes #304

## Changes
- Added CHANGELOG entry under `[0.6.5]` documenting the behavioral change from truncated-index (`int(len * quantile)`) to ceil-rank (`math.ceil(quantile * len)`) for global `p50/p95/p99_response_time_ms` keys in `get_metrics()`, introduced in PR #296.
- Added `test_percentile_uses_ceil_rank_for_short_and_even_lists()` to `tests/unit/test_monitoring.py` pinning ceil-rank behavior on empty, single-element, even-length (2 elements), and short (3–4 element) sample lists, including a case demonstrating divergence from the old formula (4 elements at p50: old=30.0, new=20.0).

## Test plan
- `uv run pytest tests/unit/test_monitoring.py -q` — all 11 tests pass including new regression test
- `uv run ruff check CHANGELOG.md tests/unit/test_monitoring.py` — clean
- Pre-existing mypy issue on `failing_func` unrelated to this change

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,11 @@ All notable changes to the Open Stocks MCP project will be documented in this fi
 
 ### Changed
 - Bumped version to 0.6.5
+- **Percentile calculation**: Global `p50/p95/p99_response_time_ms` values in
+  `get_metrics()` now use ceil-rank (`math.ceil(quantile * len)`) instead of the
+  previous truncated-index (`int(len * quantile)`) for percentile computation,
+  introduced in PR #296. Values may differ by one rank position for small or
+  even-length sample windows (e.g., 4 samples at p50: old=30.0, new=20.0).
 - Code cleanup for release (release cleanup)
 - Updated all dependencies to latest versions
 - Added open source community infrastructure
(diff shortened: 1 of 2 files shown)


unseriousAI (approved):
AGENTSHORE_CODE_REVIEW
head_sha: b6ccc987b962b2d45ce95b44d1855329af741735
status: PASS
spec_compliance: PASS
blocking_findings: 0
non_blocking_findings: 0

## Summary
APPROVE: PR cleanly documents and regression-tests the percentile behavior change requested by issue #304.

## Blocking
None

## Suggestions
None

PR #890, opened 2026-05-27
fix: prune empty deques from tool_usage and tool_response_times on cleanup

### LINKED ISSUE #306: Follow-up: prune empty deques from tool_usage and tool_response_times on cleanup

`MetricsCollector._clean_old_entries` empties the per-tool deques in `tool_usage` and `tool_response_times` once samples age out of the window, but never removes the keys themselves. In a long-running server that observes many distinct tool names over time (or in tests that exercise many tools), empty deques accumulate forever.

Action: after the inner `popleft` loop, drop keys whose deques are empty. (Or switch storage to an LRU/TTL map.) Pre-existing in `tool_usage`; introduced for `tool_response_times` in #162.

jwesleye:
Closes #306

## Changes
- After each `popleft` loop in `_clean_old_entries`, collect and delete keys whose deques are now empty from both `tool_usage` and `tool_response_times`
- Updated existing `test_get_metrics_cleans_stale_tool_samples` to assert the key is absent (not present with zero calls) since pruning now removes it entirely
- Added `test_clean_old_entries_prunes_empty_deques` verifying: stale keys are pruned, recent keys are preserved with data intact, and `record_api_call` can re-create a pruned key via `defaultdict`

## Test plan
- `uv run pytest tests/unit/test_monitoring.py tests/unit/test_monitoring_alerts.py -q --tb=line` — 23 passed
- `uv run ruff check src/open_stocks_mcp/monitoring.py tests/unit/test_monitoring.py` — no issues
- `uv run mypy src/open_stocks_mcp/monitoring.py --no-error-summary` — no errors

### DIFF
--- a/src/open_stocks_mcp/monitoring.py
+++ b/src/open_stocks_mcp/monitoring.py
@@ -197,11 +197,17 @@ class MetricsCollector:
         for tool_calls in self.tool_usage.values():
             while tool_calls and tool_calls[0][0] < cutoff:
                 tool_calls.popleft()
+        empty_tools = [t for t, d in self.tool_usage.items() if not d]
+        for t in empty_tools:
+            del self.tool_usage[t]
 
         # Clean per-tool response times
         for tool_durations in self.tool_response_times.values():
             while tool_durations and tool_durations[0][0] < cutoff:
                 tool_durations.popleft()
+        empty_tools = [t for t, d in self.tool_response_times.items() if not d]
+        for t in empty_tools:
+            del self.tool_response_times[t]
 
         while (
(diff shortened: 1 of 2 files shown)


unseriousAI (approved):
AGENTSHORE_CODE_REVIEW
head_sha: a7be46930586b3651be250964c8627dc7b0a87ec
status: PASS
spec_compliance: PASS
blocking_findings: 0
non_blocking_findings: 0

## Summary
Pass: implementation matches issue scope and safely prunes empty per-tool deques without regressions.

## Blocking
None

## Suggestions
None

PR #893, opened 2026-05-27
test: add metrics bearer bypass regression coverage

### LINKED ISSUE #308: Follow-up: add explicit test that /metrics bypasses Bearer auth when api_key is set

`tests/http_transport/test_http_server.py::TestMetricsEndpoint::test_metrics_endpoint_no_auth_and_plain_text` calls `create_http_server` without an `api_key`, so it doesn't actually exercise the no-auth contract for `/metrics` when other paths are protected. Behavior is correct by construction (`/metrics` is not in `PROTECTED_PATHS`), but a regression that silently added it to the protected set would not be caught by tests.

Action: add a test that builds the app with `api_key="secret"`, hits `/mcp` without auth (expect 401), then hits `/metrics` without auth (expect 200, Prometheus body).

unseriousAI:
Closes #308

## Changes
- add `test_metrics_bypasses_bearer_auth` in `TestBearerTokenAuth`
- verify `/mcp` returns 401 without bearer token when `api_key` is configured
- verify `/metrics` remains unauthenticated and returns Prometheus text body

## Test plan
- `uv run pytest tests/http_transport/test_http_auth.py::TestBearerTokenAuth::test_metrics_bypasses_bearer_auth -q --tb=line` (blocked by existing pyproject.toml duplicate key)
- `uv run pytest tests/http_transport/test_http_auth.py -q --tb=line` (blocked by existing pyproject.toml duplicate key)
- `uv run ruff check tests/http_transport/test_http_auth.py` (blocked by existing pyproject.toml duplicate key)

### DIFF
--- a/tests/http_transport/test_http_auth.py
+++ b/tests/http_transport/test_http_auth.py
@@ -372,6 +372,20 @@ class TestBearerTokenAuth:
         response = await authed_client.get("/")
         assert response.status_code == 200
 
+    async def test_metrics_bypasses_bearer_auth(
+        self, authed_client: httpx.AsyncClient
+    ) -> None:
+        mcp_response = await authed_client.post(
+            "/mcp",
+            json={"jsonrpc": "2.0", "method": "tools/list", "id": 1},
+        )
+        assert mcp_response.status_code == 401
+
+        metrics_response = await authed_client.get("/metrics")
+        assert metrics_response.status_code == 200
+        assert metrics_response.headers["content-type"].startswith("text/plain")
+        assert "open_stocks_mcp_total_calls" in metrics_response.text
+
 
 @pytest.mark.integration
 @pytest.mark.journey_system


jwesleye (approved):
AGENTSHORE_CODE_REVIEW
head_sha: b0c89c597c3d93ecf44afd3ba66e31f8de010c1b
status: PASS
spec_compliance: PASS
blocking_findings: 0
non_blocking_findings: 2

## Summary
Clean regression test that correctly pins the invariant: `/mcp` requires bearer auth, `/metrics` does not.

## Blocking
None

## Suggestions
1. `tests/http_transport/test_http_auth.py:378-382` — The first assertion (`mcp_response.status_code == 401`) is already covered by `test_mcp_endpoint_requires_token`. Its presence here as a contrast/precondition check is reasonable for a regression test, but a one-line comment (e.g. `# confirm auth IS enforced — makes the bypass below meaningful`) would make the intent explicit to future readers.

2. `tests/http_transport/test_http_auth.py:375-387` — Sibling methods like `test_health_remains_public` (line 364) include a docstring explaining the rationale. Adding one here (e.g. `"""Prometheus scrape endpoint must remain public when bearer auth is active."""`) would be consistent.

Both are style-only; no change needed before merge.

PR #900, opened 2026-05-27
test: simplify rate-limited nodeid count logic

### LINKED ISSUE #323: Follow-up: simplify dead-code count logic in test_rate_limited_collection_nodeids
Identified during review of PR #303.

In `tests/unit/test_rate_limited_marker.py`, `test_rate_limited_collection_nodeids` uses `-q` which always produces nodeid format output. The first branch of the count-checking logic:

```python
count = result.stdout.count("<Function") + result.stdout.count("<Coroutine")
if count == 0:   # always True with -q
    lines = [line for line in result.stdout.splitlines() if "::" in line]
    count = len(lines)
```

…is dead code; the tree-format path never executes. Consider simplifying to:

```python
count = sum(1 for line in result.stdout.splitlines() if "::" in line and not line.startswith("no tests ran"))
assert count == 4, f"Expected 4 tests, found {count}. Output:\n{result.stdout}"
```

unseriousAI:
Closes #323

## Changes
- Removed unreachable tree-format collection count path in `test_rate_limited_collection_nodeids`
- Replaced it with direct nodeid-format line counting that ignores `no tests ran` summary lines
- Kept the exact assertion (`count == 4`) and all presence/absence checks unchanged

## Test plan
- `uv run pytest tests/unit/test_rate_limited_marker.py::test_rate_limited_collection_nodeids -q --tb=line` *(blocked: duplicate `pythonpath` key in `pyproject.toml`)*
- `python3 -m pytest -c /dev/null tests/unit/test_rate_limited_marker.py::test_rate_limited_collection_nodeids -q --tb=line` *(blocked: `open_stocks_mcp` import bootstrap requirement)*
- `ruff check --isolated tests/unit/test_rate_limited_marker.py`
- `mypy --config-file /dev/null --ignore-missing-imports --follow-imports=skip tests/unit/test_rate_limited_marker.py`

### DIFF
--- a/tests/unit/test_rate_limited_marker.py
+++ b/tests/unit/test_rate_limited_marker.py
@@ -201,13 +201,11 @@ def test_rate_limited_collection_nodeids() -> None:
     assert "test_api_error_handling" not in result.stdout
 
     # Also check count to be sure no extra tests are selected
-    # Tree format usually has <Function ...> or <Coroutine ...>
-    count = result.stdout.count("<Function") + result.stdout.count("<Coroutine")
-    # If it's nodeid format, it might not have those, so we fallback
-    if count == 0:
-        # Fallback for nodeid format (one per line)
-        lines = [line for line in result.stdout.splitlines() if "::" in line]
-        count = len(lines)
+    count = sum(
+        1
+        for line in result.stdout.splitlines()
+        if "::" in line and not line.startswith("no tests ran")
+    )
 
     assert count == 4, f"Expected 4 tests, found {count}. Output:\n{result.stdout}"
 


jwesleye (approved):
AGENTSHORE_CODE_REVIEW
head_sha: 22b6ba2baed7fd07973029a2308e195e493a5c16
status: PASS
spec_compliance: PASS
blocking_findings: 0
non_blocking_findings: 1

## Summary
Clean dead-code removal — the tree-format branch (`<Function`/`<Coroutine`) was unreachable with `-q`, and the new generator expression is correct and idiomatic.

## Blocking
None

## Suggestions
1. **tests/unit/test_rate_limited_marker.py, lines 207** — the `not line.startswith("no tests ran")` guard is redundant. A line starting with `"no tests ran"` can never also contain `"::"` (pytest never produces `"no tests ran::something"`), so the `"::" in line` check already excludes it. The guard is harmless but adds noise. Consider simplifying to:
   ```python
   count = sum(1 for line in result.stdout.splitlines() if "::" in line)
   ```
   Low-priority follow-up; does not affect correctness.

PR #912, opened 2026-05-27
refactor: add _require_health_service() guard consistent with existing helpers

### LINKED ISSUE #361: Follow-up: add _require_health_service() helper consistent with existing guard pattern (from PR #328)
Identified during review of PR #328.

`src/open_stocks_mcp/server/http_transport.py` uses `_require_session_manager()` and `_require_metrics_collector()` helpers that explicitly check for None singletons and raise 503. The new `health_check` endpoint calls `get_health_service()` directly without a parallel guard. Since `get_health_service()` always creates an instance, this is not a bug, but the inconsistency means the failure mode is an implicit AttributeError rather than an explicit 503.

Consider adding:
```python
def _require_health_service() -> HealthService:
    svc = get_health_service()
    if svc is None:
        raise HTTPException(status_code=503, detail="Service unhealthy")
    return svc
```
and using it in `health_check`.

jwesleye:
Closes #361

## Changes
- Added `_require_health_service()` helper in `http_transport.py`, consistent with the existing `_require_session_manager()` and `_require_metrics_collector()` guard pattern
- Updated `health_check()` endpoint to use `_require_health_service()` instead of calling `get_health_service()` directly, making the failure mode an explicit 503 HTTPException rather than an implicit AttributeError
- Added unit test `test_health_returns_503_when_health_service_unavailable` covering the new guard

## Test plan
- `pytest tests/unit/test_http_transport.py -q` — all 16 tests pass including the new one
- `ruff check src/open_stocks_mcp/server/http_transport.py tests/unit/test_http_transport.py` — clean
- `mypy src/open_stocks_mcp/server/http_transport.py` — clean

### DIFF
--- a/src/open_stocks_mcp/server/http_transport.py
+++ b/src/open_stocks_mcp/server/http_transport.py
@@ -209,6 +209,14 @@ def _require_metrics_collector() -> Any:
     return metrics_collector
 
 
+def _require_health_service() -> Any:
+    """Return health service or raise 503 when unavailable."""
+    svc = get_health_service()
+    if svc is None:
+        raise HTTPException(status_code=503, detail="Service unhealthy")
+    return svc
+
+
 class BearerAuthMiddleware(BaseHTTPMiddleware):
     """Require ``Authorization: Bearer <api_key>`` on protected endpoints.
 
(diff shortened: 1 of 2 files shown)


unseriousAI (approved):
AGENTSHORE_CODE_REVIEW
head_sha: a269b6f58d14ed6f6af1529b119a6582ae20b17d
status: PASS
spec_compliance: PASS
blocking_findings: 0
non_blocking_findings: 0

## Summary
Approve: helper guard is consistent with existing server patterns and test coverage validates the new 503 behavior.

## Blocking
None

## Suggestions
None

PR #913, opened 2026-05-27
fix: return -32600 for missing/non-string MCP method

### LINKED ISSUE #363: Follow-up: return -32600 for missing method field in MCP JSON-RPC
Identified during review of PR #338.

In `src/open_stocks_mcp/server/http_transport.py` (the `mcp_endpoint` handler), when an incoming JSON-RPC request omits the `"method"` field entirely, `json_request.get("method")` returns `None`. This falls through all `elif` branches and hits the new `else` clause added in PR #338, returning a `-32601` (Method Not Found) response with the message `"Method not found: None"`.

Per [JSON-RPC 2.0 spec §5](https://www.jsonrpc.org/specification#error_object), a missing or non-string `method` is an **Invalid Request** (`-32600`), not Method Not Found (`-32601`).

## Fix

Add a guard before the method dispatch block:

```python
if not isinstance(method, str):
    return Response(
        content=json.dumps({
            "jsonrpc": "2.0",
            "error": {"code": -32600, "message": "Invalid Request: method missing or not a string"},
            "id": request_id,
        }).encode(),
        status_code=200,
        headers={"content-type": "application/json"},
    )
```

Also add a unit test case sending `{"jsonrpc": "2.0", "id": 1}` (no method) and asserting `error.code == -32600`.

unseriousAI:
Closes #363

## Changes
- add MCP JSON-RPC guard in `/mcp` handler to treat missing/non-string `method` as Invalid Request (`-32600`)
- preserve unknown string method behavior as Method Not Found (`-32601`)
- add unit tests for missing and non-string `method` payloads and id preservation

## Test plan
- uv run pytest tests/unit/test_http_transport.py::TestHttpTransportUnit -q --tb=line
- uv run ruff check src/open_stocks_mcp/server/http_transport.py tests/unit/test_http_transport.py

### DIFF
--- a/src/open_stocks_mcp/server/http_transport.py
+++ b/src/open_stocks_mcp/server/http_transport.py
@@ -596,6 +596,26 @@ def create_http_server(
             params = json_request.get("params", {})
             _log_mcp_request(method, request_id)
 
+            if not isinstance(method, str):
+                return _mcp_json_response(
+                    {
+                        "jsonrpc": "2.0",
+                        "error": {
+                            "code": -32600,
+                            "message": (
+                                "Invalid Request: method missing or not a string"
+                            ),
+                        },
+                        "id": request_id,
+                    },
+                    status_code=200,
+                    method=method,
+                    request_id=request_id,
+                    outcome="error",
+                    error_code=-32600,
+                    error_type="invalid_request",
+                )
+
             try:
                 result: dict[str, Any]
                 if method == "initialize":
--- a/tests/unit/test_http_transport.py
+++ b/tests/unit/test_http_transport.py
@@ -89,6 +89,29 @@ class TestHttpTransportUnit:
         assert "method not found" in body["error"]["message"].lower()
         assert body["id"] == 7
 
+    def test_mcp_missing_method_returns_invalid_request_minus_32600(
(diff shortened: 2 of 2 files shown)


jwesleye (approved):
AGENTSHORE_CODE_REVIEW
head_sha: f9475919118837e3356be8b25bd00ee563c37d55
status: PASS
spec_compliance: PASS
blocking_findings: 0
non_blocking_findings: 0

## Summary
Clean, well-scoped fix that correctly returns JSON-RPC 2.0 error code -32600 (Invalid Request) for missing or non-string `method` fields, as specified in issue #363.

## Blocking
None

## Suggestions
None

**Details reviewed:**
- `isinstance(method, str)` guard correctly catches both `None` (missing key via `dict.get`) and non-string types (int, list, etc.)
- Uses existing `_mcp_json_response` helper consistently with the rest of the handler
- `request_id` properly echoed back per JSON-RPC 2.0 spec
- Existing `-32601` (Method Not Found) for unknown string methods preserved and unaffected
- Two unit tests cover missing-method and null-method cases, both verifying error code, message content, and id preservation
- All 15 http_transport unit tests pass

PR #920, opened 2026-05-27
fix: broker comparison _has_error catches no_data status

### LINKED ISSUE #396: Follow-up: broker_comparison _has_error misses no_data status
Identified during review of PR #354.

`_has_error` in `src/open_stocks_mcp/tools/broker_comparison_tools.py` (lines 25-32) checks for `"error"` key and a set of status strings, but does not cover `"no_data"` status. `get_portfolio()` returns `create_no_data_response(...)` (status `"no_data"`) when the API returns nothing — this can happen on an auth failure that doesn't raise. The result is that Robinhood renders as `available: True` with all-zero summary fields instead of `available: False`.

Fix: add `"no_data"` to the status set in `_has_error`.

Note: the same gap exists in `cross_broker_tools._collect_robinhood_data` — that function uses `"error" not in portfolio_data` which also misses `no_data`.

jwesleye:
Closes #396

## Changes
- Add `_has_error(data)` helper to `broker_comparison_tools.py` that checks for both `"error"` key and `"no_data"` status
- Replace all inline `"error" not in data` checks in both `_collect_robinhood_comparison` and `_collect_schwab_comparison` with the new helper
- When `get_portfolio()` returns `no_data`, Robinhood comparison now correctly sets `available: False` with a descriptive note instead of silently showing `available: True` with all-zero summary fields

## Test plan
- Added `test_robinhood_no_data_portfolio_marks_unavailable` — verifies `available=False` and notes populated when portfolio returns `no_data`
- Added `test_no_data_pricing_excluded_from_comparison` — verifies `no_data` pricing responses are excluded from comparison output
- Existing tests continue to pass (4/4 in broker comparison, 11/11 in cross broker)

### DIFF
--- a/src/open_stocks_mcp/tools/broker_comparison_tools.py
+++ b/src/open_stocks_mcp/tools/broker_comparison_tools.py
@@ -25,6 +25,14 @@ def _safe_float(value: Any, default: float = 0.0) -> float:
         return default
 
 
+_ERROR_STATUSES = {"error", "no_data"}
+
+
+def _has_error(data: dict[str, Any]) -> bool:
+    """Return True if the response dict indicates an error or missing data."""
+    return "error" in data or data.get("status") in _ERROR_STATUSES
+
+
 async def _collect_robinhood_comparison(
     symbols: list[str] | None = None,
     include_orders: bool = True,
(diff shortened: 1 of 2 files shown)


unseriousAI (approved):
AGENTSHORE_CODE_REVIEW
head_sha: 7539153944ff76ee5657e08618ce0ade81d889c1
status: PASS
spec_compliance: PASS
blocking_findings: 0
non_blocking_findings: 0

## Summary
Approved: fix correctly handles no_data status in broker comparison paths with targeted regression coverage.

## Blocking
None

## Suggestions
None

PR #921, opened 2026-05-27
fix: preserve registry order in get_aggregated_portfolio brokers dict

### LINKED ISSUE #542: Follow-up: Preserve registry order in get_aggregated_portfolio brokers dict
Identified during review of PR #538.

`src/open_stocks_mcp/tools/cross_broker_tools.py:122-208` — after the parallelization rewrite, the `brokers` field in the returned payload no longer matches `registry.list_brokers()` ordering. The new implementation inserts all available brokers first (in registry order) and then appends unavailable brokers (in registry order), instead of interleaving them in `broker_names` order as before.

The PR description claims "preserved stable broker output ordering from `registry.list_brokers()`", and the underlying acceptance criterion in #535 calls for `list_brokers()` ordering, so the implementation drifts from intent.

Suggested fix:

```python
results_map = dict(zip(available_names, results, strict=True))
for name in broker_names:
    if name in unavailable_records:
        brokers_out[name] = unavailable_records[name]
        continue
    result = results_map[name]
    # existing per-result processing...
```

No functional/data regression — only key ordering in the output dict.

jwesleye:
Closes #542

## Changes
- Refactored `get_aggregated_portfolio` to build `results_map` after `asyncio.gather`, then iterate `broker_names` in a single pass to construct `brokers_out` in registry order
- Removed the two-phase approach (available-first, then unavailable) that caused insertion-order drift
- Added a regression test with three brokers where the middle one is unavailable, asserting dict key order matches `registry.list_brokers()`

## Test plan
- `pytest tests/unit/test_cross_broker_tools.py` — all 12 tests pass including new ordering test
- `pytest -m journey_portfolio` — 24 passed, 3 skipped
- `mypy src/open_stocks_mcp/tools/cross_broker_tools.py` — no issues
- `ruff check src/open_stocks_mcp/tools/cross_broker_tools.py` — no issues

### DIFF
--- a/src/open_stocks_mcp/tools/cross_broker_tools.py
+++ b/src/open_stocks_mcp/tools/cross_broker_tools.py
@@ -59,50 +59,55 @@ async def get_aggregated_portfolio() -> dict[str, Any]:
         available_names.append(name)
         coros.append(broker.get_portfolio_snapshot())
 
+    results_map: dict[str, Any] = {}
     if coros:
         results = await asyncio.gather(*coros, return_exceptions=True)
-        for name, result in zip(available_names, results, strict=True):
-            if isinstance(result, BaseException):
-                logger.error(
-                    f"Error collecting data from broker {name}: {result}",
-                    exc_info=result,
-                )
-                brokers_out[name] = {
-                    "status": "error",
-                    "error": str(result),
-                    "market_value": 0.0,
-                    "equity": 0.0,
-                    "buying_power": 0.0,
-                    "positions": [],
-                    "position_count": 0,
-                }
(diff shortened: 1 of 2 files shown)


unseriousAI (approved):
AGENTSHORE_CODE_REVIEW
head_sha: 63c7b7299e20e2d300e9ac56e20789e8b3886805
status: PASS
spec_compliance: PASS
blocking_findings: 0
non_blocking_findings: 0

## Summary
Meets issue #542: brokers map now preserves registry order while retaining concurrent snapshot collection and existing degraded/error semantics.

## Blocking
None

## Suggestions
None

PR #925, opened 2026-05-27
fix: remove duplicate market rate-limit acquires

### LINKED ISSUE #872: Follow-up: migrate remaining market tools to broker-specific rate limiter
Identified during review of PR #774.

Several Robinhood market tools still call `get_rate_limiter()` without a broker name, using the global limiter instead of the broker-specific limiter introduced in #774. These tools also delegate to `execute_with_retry` which now acquires against the broker-specific limiter, creating redundant rate-limit acquisition against two independent instances.

**Affected files:**
- `src/open_stocks_mcp/tools/market/movers.py` (4 call sites)
- `src/open_stocks_mcp/tools/market/earnings.py` (3 call sites)
- `src/open_stocks_mcp/tools/market/news.py` (1 call site)
- `src/open_stocks_mcp/tools/market/ratings.py` (1 call site)
- `src/open_stocks_mcp/tools/market/level2.py` (1 call site)
- `src/open_stocks_mcp/server/tool_helpers.py` (1 call site)

**Options:**
1. Migrate direct calls to `get_rate_limiter("robinhood")`
2. Remove the redundant manual `acquire()` calls since `execute_with_retry` already rate-limits internally

unseriousAI:
Closes #872

## Changes
- removed direct `get_rate_limiter().acquire()` calls from Robinhood market tools that already use `execute_with_retry`
- switched `get_rate_limit_status_data` to read Robinhood broker-specific limiter via `get_rate_limiter("robinhood")`
- updated unit tests to stop mocking removed manual limiter calls and to assert broker-specific limiter selection in server helper tests

## Test plan
- `pytest -q -o addopts='' tests/unit/test_market_research_tools.py`
- `pytest -q -o addopts='' tests/unit/test_market_research_tools.py tests/server/test_tool_helpers.py` (fails in this environment during collection due to missing optional dependency: `ModuleNotFoundError: No module named "schwab"`)

### DIFF
--- a/src/open_stocks_mcp/server/tool_helpers.py
+++ b/src/open_stocks_mcp/server/tool_helpers.py
@@ -117,7 +117,7 @@ async def get_list_brokers_data() -> dict[str, Any]:
 
 async def get_rate_limit_status_data() -> dict[str, Any]:
     """Return current rate limit usage and statistics."""
-    rate_limiter = get_rate_limiter()
+    rate_limiter = get_rate_limiter("robinhood")
     stats = rate_limiter.get_stats()
 
     return {
--- a/src/open_stocks_mcp/tools/market/earnings.py
+++ b/src/open_stocks_mcp/tools/market/earnings.py
@@ -15,7 +15,6 @@ from open_stocks_mcp.tools.error_handling import (
     log_api_call,
     validate_symbol,
 )
-from open_stocks_mcp.tools.rate_limiter import get_rate_limiter
 
 
 @handle_robin_stocks_errors
@@ -67,10 +66,6 @@ async def get_stock_earnings(symbol: str) -> dict[str, Any]:
             return create_error_response(
                 ValueError("Authentication required"), "authentication"
             )
-
-        # Apply rate limiting
-        rate_limiter = get_rate_limiter()
-        await rate_limiter.acquire()
 
         log_api_call("get_stock_earnings", symbol=symbol)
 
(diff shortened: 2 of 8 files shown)


jwesleye (approved):
AGENTSHORE_CODE_REVIEW
head_sha: f0dc63ccc530da18447a42c50c4d1148d80c867c
status: PASS
spec_compliance: PASS
blocking_findings: 0
non_blocking_findings: 2

## Summary

Clean removal of 10 duplicate `rate_limiter.acquire()` calls from market tools that already delegated to `execute_with_retry`, plus a correct fix to `get_rate_limit_status_data` to report the broker-registry limiter rather than the disconnected global instance.

## Blocking

None

## Suggestions

1. **`tool_helpers.py:120` — unhandled `RegistryNotInitializedError` in `get_rate_limit_status_data`**
   `get_rate_limiter("robinhood")` now calls `get_broker_registry_sync()`, which raises `RegistryNotInitializedError` if `get_broker_registry()` has not yet been awaited during startup. The function has no try/except, unlike sibling helpers (`get_broker_status_data`, `get_list_brokers_data`). On a healthy running server this is fine; an early-startup health check would surface an unhandled exception rather than graceful error stats. Suggest wrapping with a try/except that falls back to the global legacy limiter or returns an error-shape response. Follow-up issue: #926

2. **`robinhood_dividend_tools.py:54–399` — 5 tools still use manual `acquire()` without `execute_with_retry`**
   These are NOT duplicates (no `execute_with_retry` is present), so they're correctly out of scope for this PR. But they miss circuit-breaker protection, retry logic, and request coordination that `execute_with_retry` provides. Migrating them would align them with the rest of the Robinhood tool layer. Follow-up issue: #927

PR #943, opened 2026-05-27
fix: handle uninitialized registry in rate-limit status

### LINKED ISSUE #937: Follow-up: add error handling to get_rate_limit_status_data for uninitialized registry
Identified during review of PR #925.

`get_rate_limit_status_data` in `src/open_stocks_mcp/server/tool_helpers.py` was changed to call `get_rate_limiter("robinhood")`, which internally calls `get_broker_registry_sync()`. If the async broker registry hasn't been initialized yet (e.g., an early-startup health check before the server's `lifespan` hook runs), this raises `RegistryNotInitializedError` with no error handling.

Sibling helpers (`get_broker_status_data`, `get_list_brokers_data`) already wrap their registry calls in try/except. This function should do the same — either falling back to the legacy global limiter or returning an error-shape response.

**Suggested fix:** wrap the `get_rate_limiter("robinhood")` call in a try/except `RegistryNotInitializedError` block and fall back to `get_rate_limiter()` (the global lazy limiter) with a note in the returned stats.

unseriousAI:
Closes #937

## Changes
- catch `RegistryNotInitializedError` in `get_rate_limit_status_data` when requesting the Robinhood-scoped limiter
- fall back to the global lazy rate limiter during early startup before broker registry initialization
- include a `note` field in the response when fallback is used
- add server helper test coverage for the fallback path

## Test plan
- uv run pytest tests/server/test_tool_helpers.py -q
- uv run ruff check src/open_stocks_mcp/server/tool_helpers.py tests/server/test_tool_helpers.py
- uv run mypy src/open_stocks_mcp/server/tool_helpers.py

### DIFF
--- a/src/open_stocks_mcp/server/tool_helpers.py
+++ b/src/open_stocks_mcp/server/tool_helpers.py
@@ -9,7 +9,10 @@ from typing import Any
 
 from mcp.server.fastmcp import FastMCP
 
-from open_stocks_mcp.brokers.registry import get_broker_registry
+from open_stocks_mcp.brokers.registry import (
+    RegistryNotInitializedError,
+    get_broker_registry,
+)
 from open_stocks_mcp.brokers.session_state import get_session_manager
 from open_stocks_mcp.health import get_health_service
 from open_stocks_mcp.logging_config import logger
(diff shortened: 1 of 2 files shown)


jwesleye (approved):
AGENTSHORE_CODE_REVIEW
head_sha: dbdaf5ab233baa7a9f680fe90ce1d98dbd86f85d
status: PASS
spec_compliance: PASS
blocking_findings: 0
non_blocking_findings: 0

## Summary
Clean, minimal fix that exactly addresses #937. Catches `RegistryNotInitializedError` during early startup and falls back to the global lazy rate limiter with a descriptive `note` field. Test coverage for the fallback path is solid.

## Blocking
None

## Suggestions
None

PR #1006, opened 2026-05-28
refactor: replace shared global rate limiter with per-broker instances

### LINKED ISSUE #990: Brokers: rate limiter shared globally; design called for per-broker
The current rate limiter implementation is shared globally across all brokers. The original multi-broker design called for per-broker rate limiting to allow independent configuration and prevent one broker's rate limits from affecting another broker's throughput. This needs to be refactored so each broker (Robinhood, Schwab) has its own rate limiter instance with broker-specific settings.

jwesleye:
Closes #990

## Changes
- Removed `configure_global_rate_limiter()` call from `create_mcp_server()` in `server/app.py` — the per-broker registry already creates instances on demand with correct defaults; the global was never used for actual API calls
- Removed global `get_rate_limiter()` warmup from HTTP transport lifespan (no-op for real throttling)
- Updated `GET /status` endpoint to report per-broker rate limiter stats (`rate_limiting.robinhood.*` and `rate_limiting.schwab.*`) instead of the shared global counter
- Updated `rate_limit_status` MCP tool (`get_rate_limit_status_data`) to collect stats from `BrokerRegistry.get_rate_limiter(broker_name)` for both brokers, with a fallback that shows default limits when the registry hasn't initialized yet
- Updated all affected tests to reflect the new per-broker response shape

## Why
The old `configure_global_rate_limiter()` path wired up a single `RateLimiter` initialized with Robinhood's config. The `/status` endpoint and `rate_limit_status` MCP tool read from this global — so Schwab call stats were invisible, and operators had no per-broker visibility. Actual API calls already went through the correct per-broker path (`BrokerRegistry.get_rate_limiter`); this change removes the misleading parallel global path and surfaces real per-broker data.

## Test plan
- `pytest tests/unit/test_config.py tests/unit/test_simple_rate_limiter.py tests/unit/test_rate_limiter_isolation.py tests/server/test_server_app.py` — all pass
- `pytest -m "not slow and not exception_test and not rate_limited and not performance and not live_market"` — 1592 passed, 0 failed

### DIFF
--- a/src/open_stocks_mcp/server/app.py
+++ b/src/open_stocks_mcp/server/app.py
@@ -55,7 +55,6 @@ from open_stocks_mcp.tools.options.market_data import (
     get_option_historicals,
     get_option_market_data,
 )
-from open_stocks_mcp.tools.rate_limiter import configure_global_rate_limiter
 from open_stocks_mcp.tools.robinhood_account_feature_summary_tools import (
     get_account_features,
 )
@@ -2034,11 +2033,6 @@ def create_mcp_server(config: ServerConfig | None = None) -> FastMCP:
         config = load_config()
 
     setup_logging(config)
-    configure_global_rate_limiter(
-        config.rate_limits.calls_per_minute,
-        config.rate_limits.calls_per_hour,
-        config.rate_limits.burst_size,
-    )
     setup_tracing(config)
     instrument_mcp_tool_calls(mcp)
     if config.timeout is not None:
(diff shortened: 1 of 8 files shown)


unseriousAI (approved):
AGENTSHORE_CODE_REVIEW
head_sha: 2f36aec5a13423010018c430c0e84042ea4fdd40
status: PASS
spec_compliance: PASS
blocking_findings: 0
non_blocking_findings: 0

## Summary
Per-broker rate limiter refactor matches issue scope and implementation is consistent across server status/tool helper surfaces and tests.

## Blocking
None

## Suggestions
None

